PR #1, opened 2024-05-05
Subscription Management

Sjmarf:
See [this PR](https://github.com/mlemgroup/mlem/pull/1052)


Sources/MlemMiddleware/Content Models/SubscriptionList/SubscriptionList.swift
```diff
@@ -0,0 +1,102 @@
...
+            self.alphabeticSections[section.key] = section.value.sorted(by: { $0.name < $1.name })
+        }
+    }
+    
+    func updateCommunitySubscription(community: Community2) {
+        if community.subscribed {
+            if !self.communities.contains(community) {
+                self.addCommunity(community: community)
+            }
+            if self.favoriteIDs.contains(community.id) != community.shouldBeFavorited {
```
EricBAndrews: `self.favoriteIDs.contains(communityId)` is probably code we're going to write a lot--might be a little prettier to extract to `SubscriptionList.isFavorited(communityId)`

Sources/MlemMiddleware/Content Models/SubscriptionList/SubscriptionList.swift
```diff
@@ -0,0 +1,102 @@
...
+                if community.shouldBeFavorited {
+                    self.favoriteIDs.insert(community.id)
+                    let index = self.favorites.insertionIndex { $0.name < community.name }
+                    self.favorites.insert(community, at: index)
```
EricBAndrews: This same logic is duplicated below, should probably be moved into `Arrays+Extensions` as something like `Array.sortedInsert(item, condition)`

Sources/MlemMiddleware/Content Models/SubscriptionList/SubscriptionList.swift
```diff
@@ -0,0 +1,102 @@
...
+                } else {
+                    self.favoriteIDs.remove(community.id)
+                    if let index = favorites.firstIndex(of: community) {
+                        favorites.remove(at: index)
+                    }
```
EricBAndrews: Is there a reason not to just use `.removeFirst(community)`?

Sources/MlemMiddleware/Content Models/SubscriptionList/SubscriptionList.swift
```diff
@@ -0,0 +1,102 @@
...
+        self.communities.remove(community)
+        self.favoriteIDs.remove(community.id)
+        if let index = self.favorites.firstIndex(of: community) {
+            self.favorites.remove(at: index)
+        }
+        let category = self.categoryForCommunity(community)
+        if let index = self.alphabeticSections[category]?.firstIndex(of: community) {
+            self.alphabeticSections[category]?.remove(at: index)
+        }
+    }
```
EricBAndrews: Ditto above comments re: remove and sorted insert

EricBAndrews (changes requested):
Looks good! A couple minor comments about some array operations.

Sources/MlemMiddleware/Content Models/SubscriptionList/SubscriptionList.swift
```diff
@@ -0,0 +1,102 @@
...
+    
+    var favoriteIDs: Set<Int> {
+        get { getFavorites() }
+        set { self.setFavorites(newValue) }
+    }
+    
+    private var getFavorites: () -> Set<Int>
+    private var setFavorites: (Set<Int>) -> Void
+    
+    private var apiClient: ApiClient
```
EricBAndrews: Should just be `api` for consistency with other models

Sources/MlemMiddleware/Content Models/SubscriptionList/SubscriptionList.swift
Sjmarf: Unfortunately there isn't a better way to do this, unless we define one in an extension. `removeFirst(n)`  removes the first `n` elements from the array.

EricBAndrews: Oh, right, forgot that's what that one does 🤦‍♂️

Probably worth defining in an extension, then.

Sources/MlemMiddleware/Content Models/SubscriptionList/SubscriptionList.swift
Sjmarf: Makes sense 👍

EricBAndrews (approved):
LGTM 👍 

`apiClient` in `SubscriptionList` should still be `api`, small enough nit that I'm not going to run us through another review cycle for it lol

PR #6, opened 2024-05-12
`Instance` and `UserStub` changes

Sjmarf:
This PR contains changes to support the login system I'm working on.

- The keychain identifier used for storing account tokens now uses `actorId` (which is always unique) rather than `id` (which isn't guaranteed to be unique). It's backwards-compatible with the `id` system.
- Added an `updateToken` method to `ApiClient`. This will only be used when refreshing the token for the same account, and should never be used to change the account an `ApiClient` is logged into.
- Added a `deleteTokenFromKeychain` method to `UserStub`, which I'll be calling when a user signs out.
- Conformed `Instance` models to `ActorIdentifiable`.

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -19,7 +19,7 @@ public class ApiClient {
     // url and token MAY NOT be modified! Downstream code expects that a given ApiClient will *always* submit requests from the same user to the same instance.
     public let baseUrl: URL
     let endpointUrl: URL
-    let token: String?
+    internal private(set) var token: String?
     private(set) var fetchedVersion: SiteVersion?
     private var fetchSiteTask: Task<SiteVersion, Error>?
     
@@ -61,6 +61,17 @@ public class ApiClient {
     public func cleanCaches() {
         caches.clean()
         ApiClient.apiClientCache.clean()
+    }
+    
+    /// This should **only** be used when we get a new token for **the same** account!
+    public func updateToken(_ newToken: String) {
+        guard token != nil else {
+            assertionFailure()
+            return
+        }
+        Self.apiClientCache.changeToken(for: baseUrl, oldToken: token, newToken: newToken)
+        self.token = newToken
+        self.myUser?.stub.deleteTokenFromKeychain()
     }
     
     /// Creates or retrieves an API client for the given connection parameters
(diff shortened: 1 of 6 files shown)


Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -198,21 +206,28 @@ extension ApiClient: ActorIdentifiable {
...
             
             let ret: ApiClient = .init(baseUrl: baseUrl, token: token)
             cachedItems[ret.cacheId] = .init(content: ret)
             return ret
         }
+        
+        // Should ONLY be used when we get a new token for THE SAME account
+        func changeToken(for baseUrl: URL, oldToken: String?, newToken: String?) {
+            let oldCacheId = getCacheId(for: baseUrl, with: oldToken)
+            let newCacheId = getCacheId(for: baseUrl, with: oldToken)
```
EricBAndrews: Both generate with `oldToken`--should it be

```suggestion
            let oldCacheId = getCacheId(for: baseUrl, with: oldToken)
            let newCacheId = getCacheId(for: baseUrl, with: newToken)
```
?

Sources/MlemMiddleware/Content Models/User/UserStub.swift
```diff
@@ -103,9 +100,30 @@ public final class UserStub: UserProviding, Codable {
         try container.encode(lastLoggedIn, forKey: .lastUsed)
         try container.encode(api.baseUrl, forKey: .instanceLink)
     }
+    
+    public var keychainId: String {
+        getKeychainId(actorId: actorId)
+    }
```
EricBAndrews, line 106: This feels redundant with `getKeychainId` itself--can we just define this as `"\(actorId.absoluteString)_accessToken"`? We can then do the same thing and define a `legacyKeychainId` that performs the int-based keychain id creation

EricBAndrews (approved):
Looks good, a couple minor questions

EricBAndrews (changes requested):
Oops, meant to request changes lol

Sources/MlemMiddleware/API Client/ApiClient.swift
Sjmarf: Yep, nice catch!

Sources/MlemMiddleware/Content Models/User/UserStub.swift
Sjmarf: `getKeychainId` needs to exist outside of the class because it is called in the `init(from decoder: Decoder)` initialiser - if it were inside the class, it would be inaccessible until all properties are initialised.

EricBAndrews: Makes sense 👍

EricBAndrews (approved):
🚀  LGTM

PR #7, opened 2024-05-14
AnyPost, Upgradable

EricBAndrews:
This PR introduces the `Upgradable` protocol and the `AnyPost` class as an illustration thereof.

## `Upgradable`

This is a protocol for things that can be upgraded, intended to be applied to the wrapper classes `AnyPost` and (TODO) `AnyPerson`, `AnyCommunity`, etc.

The protocol contains:
- `Base`: type of the base member of the family
- `MinimumRenderable`: lowest type in the family with sufficient content to render something meaningful to the user
- `Upgraded`: most upgraded type in the family
- `wrappedValue`: the content itself, guaranteed to be nothing more than `Base`
- `upgrade()`: function that upgrades the wrapped value to `Upgraded`
- `isUpgraded`: bool tracking whether the model is `Upgraded`

It is used primarily by the `ContentLoader`; see [this PR](https://github.com/mlemgroup/mlem/pull/1063) for details.

## Post Improvements

This PR also introduces a collection of related `Post` improvements:
- Adds `AnyPost`: this is a `Hashable, Upgradable` wrapper for `any PostStubProviding`, used in navigation and to provide `Upgradable` functionality to posts
- Adds `getPost` methods to `ApiClient`
- Adds `upgrade` methods to `PostStubProviding` and `Post1Providing`

This PR also changes `getCommunity(actorId)` to use `ResolveObject` instead of `Search`


Sources/MlemMiddleware/Content Models/Post/PostStubProviding.swift
```diff
@@ -64,3 +64,12 @@ public extension PostStubProviding {
     var read_: Bool? { nil }
     var myVote_: ScoringOperation? { nil }
 }
+
+public extension PostStubProviding {
+    func upgrade() async throws -> Post2 {
```
Sjmarf, line 69: In the community upgrade system, `upgrade` is overridden in `Community3` to simply return `self`:

```swift
// Community3.swift, line 36
public func upgrade() async throws -> Community3 { self }
```
 This avoids making requests when calling `upgrade` on an already upgraded model. For the override to work, `upgrade()` is explicitly required in the `CommunityStubProviding` definition.
 
 This implementation should be mirrored for the post upgrade system.

Sjmarf (changes requested):
Looking good! 😄 One note about consistency and optimization

Sjmarf (approved):
LGTM 👍 Remember to create a new version tag

PR #8, opened 2024-05-16
Fix `localhost`

Sjmarf:
When `ApiClient` fetches its own `Instance`, it sets a property `local` of that instance to `true`. The `guestApi` computed property of `Instance` can then be used to get the correct `ApiClient`. I've also added a couple of convenience methods to `ApiClient`, and conformed `ProfileProviding` to `ActorIdentifiable` (which is unrelated, but I'm throwing it in as I've got a hunch I'll need it later). 

The Login PR on `Mlem` is up-to-date with these changes

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -63,6 +63,16 @@ public class ApiClient {
         ApiClient.apiClientCache.clean()
     }
     
+    /// Return a new `ApiClient` withou a token.
+    public func loggedOut() -> ApiClient {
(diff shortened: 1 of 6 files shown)


Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -63,6 +63,16 @@ public class ApiClient {
         ApiClient.apiClientCache.clean()
     }
     
+    /// Return a new `ApiClient` withou a token.
```
EricBAndrews:
```suggestion
    /// Return a new `ApiClient` without a token.
```

EricBAndrews (approved):
Nice! Very clean.

PR #11, opened 2024-05-25
Mark-read Batching

### LINKED ISSUE #3: `MarkReadBatcher`
We need to rebuild this in 2.0. It should probably be attached to `ApiClient`, so that it can be accessed by posts belonging to the `ApiClient`. `Post.markAsRead()` would append the post ID to the batcher. The method could have a `commit: Bool` parameter - if `true`, it would flush the batcher after adding its own ID. Additionally, we would flush the batcher every n seconds, possibly on the same clock that we use for clearing the content model caches.

For detecting when a view goes off-screen in the `Mlem` repo, I think we might be able to use the iOS 17 `.scrollTransition` modifier for more granular control over when a post is marked as read than `.onDisappear`.

Sjmarf:
Closes #3

Added an `updateRead(_ newValue: Bool, shouldQueue: Bool)` method to `Post2Providing`. If `shouldQueue` is `true`, it will add the post ID to a `Set<Int>` stored in the `ApiClient` rather than sending an immediate request. `ApiClient.flushPostReadQueue` can be called to mark the queued posts as read. If a regular mark-read call is performed, the `ApiClient` will mark the queued posts as read in the same request.

I've switched the `cacheId` values from `actorId.hashvalue` to `id`. This allows the `ApiClient` to easily access a post with a given ID, which it uses to update the read status of posts after a mark-read request finishes.

I've also added some static properties to `SiteVersion`:

```swift
public static let v18_0: Self = .init("0.18.0")
public static let v18_1: Self = .init("0.18.1")
public static let v18_2: Self = .init("0.18.2")
public static let v18_3: Self = .init("0.18.3")
public static let v18_4: Self = .init("0.18.4")
public static let v18_5: Self = .init("0.18.5")
public static let v19_0: Self = .init("0.19.0")
public static let v19_1: Self = .init("0.19.1")
public static let v19_2: Self = .init("0.19.2")
public static let v19_3: Self = .init("0.19.3")
public static let v19_4: Self = .init("0.19.4")
 ```
 
 This allows us to compare versions like this, which I think is cleaner:
 
 ```swift
 if version >= .v19_0 {
 }
 ```
 
Side note: Do we want to name `SiteVersion` something else instead? We standardised our use of "Person" over "User", but we still mix "Instance" and "Site" in public struct/class names. IMO we should either rename `Instance` models to `Site`, or rename `SiteVersion` to `InstanceVersion`.

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
@@ -93,7 +93,7 @@ extension ApiClient: PostFeedProvider {
             if !response.success {
                 throw ApiClientError.unsuccessful
             }
-            markReadQueue.remove(id)
+            await markReadQueue.remove(id)
             if let post = caches.post2.retrieveModel(cacheId: id) {
                 post.readManager.updateWithReceivedValue(read, semaphore: semaphore)
                 post.readQueued = false
@@ -115,8 +115,7 @@ extension ApiClient: PostFeedProvider {
         let idsToSend: Set<Int>
         let markReadQueueCopy: Set<Int>
         if read, includeQueuedPosts {
-            markReadQueueCopy = self.markReadQueue
-            self.markReadQueue.removeAll()
+            markReadQueueCopy = await markReadQueue.popAll()
             idsToSend = ids.union(markReadQueueCopy)
         } else {
             markReadQueueCopy = []
@@ -132,10 +131,10 @@ extension ApiClient: PostFeedProvider {
                 throw ApiClientError.unsuccessful
             }
             if read {
-                markReadQueue.subtract(ids)
+                await markReadQueue.subtract(ids)
             }
         } catch {
-            self.markReadQueue.formUnion(markReadQueueCopy)
+            await self.markReadQueue.union(markReadQueueCopy)
             throw error
         }
         for post in idsToSend.compactMap({ caches.post2.retrieveModel(cacheId: $0) }) {
@@ -145,7 +144,7 @@ extension ApiClient: PostFeedProvider {
     }
     
     public func flushPostReadQueue() async throws {
-        if !markReadQueue.isEmpty {
+        if await !markReadQueue.ids.isEmpty {
             try await markPostsAsRead(ids: [], read: true)
         }
     }
--- a/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
@@ -77,6 +77,7 @@ extension ApiClient: PostFeedProvider {
         semaphore: UInt? = nil
     ) async throws {
         // We *must* use `postId` in 0.18 versions, and we *must* use `postIds` from 0.19.4 onwards.
+        // On versions 0.19.0 to 0.19.3, either parameter is allowed.
         let version = try await version
         let request: MarkPostAsReadRequest
         if version >= .v19_0 {
@@ -130,7 +131,9 @@ extension ApiClient: PostFeedProvider {
             if !response.success {
                 throw ApiClientError.unsuccessful
             }
-            markReadQueue.subtract(ids)
+            if read {
+                markReadQueue.subtract(ids)
+            }
         } catch {
             self.markReadQueue.formUnion(markReadQueueCopy)
             throw error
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

File: Sources/MlemMiddleware/API Client/ApiClient+Post.swift [hunk 1]
```diff
@@ -68,15 +68,94 @@ extension ApiClient: PostFeedProvider {
...
+    public func markPostAsRead(
+        id: Int,
+        read: Bool = true,
+        includeQueuedPosts: Bool = true,
+        semaphore: UInt? = nil
+    ) async throws {
+        // We *must* use `postId` in 0.18 versions, and we *must* use `postIds` from 0.19.4 onwards.
+        let version = try await version
+        let request: MarkPostAsReadRequest
+        if version >= .v19_0 {
```
EricBAndrews, line 83: Version inconsistent with comment on L79

Sources/MlemMiddleware/API Client/ApiClient+Post.swift
```diff
@@ -68,15 +68,94 @@ extension ApiClient: PostFeedProvider {
...
+        
+        guard !idsToSend.isEmpty else { return }
+        
+        do {
+            let request = MarkPostAsReadRequest(postId: nil, read: read, postIds: Array(idsToSend))
+            let response = try await perform(request)
+            if !response.success {
+                throw ApiClientError.unsuccessful
+            }
+            markReadQueue.subtract(ids)
```
EricBAndrews: Why is this line necessary? Appears duplicate with L119

Sources/MlemMiddleware/API Client/ApiClient+Post.swift
```diff
@@ -68,15 +68,94 @@ extension ApiClient: PostFeedProvider {
...
+    ) async throws {
+        let version = try await version
+        guard version >= .v19_0 else { throw ApiClientError.unsupportedLemmyVersion }
+        
+        let idsToSend: Set<Int>
+        let markReadQueueCopy: Set<Int>
+        if read, includeQueuedPosts {
+            idsToSend = ids.union(self.markReadQueue)
+            markReadQueueCopy = self.markReadQueue
+            self.markReadQueue.removeAll()
```
EricBAndrews: Not thread safe; if two threads execute L118 before either executes L119, the same request will be sent twice--hardly a cardinal sin, but could potentially result in server spam if
- `markReadQueue` has gotten long, leading to L118 executing slowly
- User is scrolling fast with mark read on scroll activated

This function should either be documented as thread-unsafe or perform thread safety checks internally.

EricBAndrews (changes requested):
Looks good! A couple notes about thread safety.

Sources/MlemMiddleware/API Client/ApiClient+Post.swift
Sjmarf: Between 0.19.0 and 0.19.4, _either_ `postId` or `postIds` can be used. I've opted to use `postIds` as soon as it's possible to do so. I'll update the comment to make this more clear 👍

Sources/MlemMiddleware/API Client/ApiClient+Post.swift
Sjmarf: When `includeQueuedPosts` is `false`, line 119 won't run. This line isn't strictly necessary, but will remove the IDs from the queue if they are present, since we just marked them as read.

Sources/MlemMiddleware/API Client/ApiClient+Post.swift
Sjmarf: Would this be enough to fix it? Or do you think I should semaphore the entire method?

```swift
markReadQueueCopy = self.markReadQueue
self.markReadQueue.removeAll()
idsToSend = ids.union(markReadQueueCopy)
```

EricBAndrews: Given just how easy it is to spin up a ton of threads all touching the underlying set (i.e., scroll fast with mark read on scroll), it's almost certain that we're going to create at best a data race and at worst a segfault unless we make sure that all of the data access to the underlying set is thread-safe.

What we should use--and, I realize now, what I should have been using all along instead of mucking around with semaphores like a C nerd--is an `actor`! Good ol' hackingwithswift has a [nice article](https://www.hackingwithswift.com/quick-start/concurrency/what-is-an-actor-and-why-does-swift-have-them) on them; as far as I can tell, the tl;dr is that everything inside an actor is executed on a serial queue, so we can just do something like:

```
actor MarkReadQueue {
  Set<Int> queuedPosts

  // clears queuedPosts and returns a set containing all removed posts
  func popAll() -> Set<Int> { ... }
}
```

Sjmarf: I didn't know about those! Thanks :D

EricBAndrews (approved):
LGTM! 🚀

PR #12, opened 2024-05-26
Remove Link Parsing

Sjmarf:
I've re-implemented this logic in LemmyMarkdownUI (Explanation [here](https://github.com/mlemgroup/mlem/issues/1072)). I don't think we should cache the links in the content models - it should be pretty easy to cache them in `@State` properties view-side instead.


EricBAndrews (approved):
👍

PR #14, opened 2024-05-30
Instance improvements

Sjmarf:
- Fleshed out instance properties (this is the majority of the line count)
- Added instance upgrade methods to `InstanceStub`.
- Added a `moderates` method to `Person3Providing` as requested [here](https://github.com/mlemgroup/mlem/pull/1071#discussion_r1617993798)
- Added `isText`, `isImage` and `isLink` to `PostType`, so we don't have to deal with `if case` in Mlem
- Renamed the `postType` property to just `type` (this is breaking and will require changes in Mlem)


EricBAndrews (approved):
LGTM 🚀

PR #17, opened 2024-06-05
New Filtering

EricBAndrews:
This PR introduces new filtering logic that unifies both deduping behavior and keyword/read filtering.

The core element is a `FilterProviding`, which is a protocol to which individual filters conform. `FilterProviding` offers `filter`, which filters a list of items and returns the ones that pass, and `reset`, which resets the filter and, if passed a list of items, filters them and returns the ones that pass.

Individual filters are then defined with whatever internal helpers they need; e.g., `PostDedupeFilter` stores a list of seen `URL`s, and is responsible for maintaining that list.

Filters can be composed together into a `MultiFilter`. A `MultiFilter` offers the same `filter` and `reset` methods, but contains a list of `FilterProviding`; `filter` calls these filters in order and returns the values that pass all filters.

Filters can also be activated or deactivated using their `active` property, and the `MultiFilter` exposes methods to activate and deactivate its internal filters. To enable this, a new `Filterable` protocol has been defined. It currently contains nothing but an `associatedtype: Filters`, which links filter types to the filter target.

`StandardFeedLoader` has been updated to contain a `MultiFilter` for its `FeedLoadable`; this is required in order for the `clearHelper` to work properly. Actually filtering posts is the responsibility of the instantiating class, as that work needs to happen down in `fetchPage`/`fetchCursor` (e.g., to allow image prefetching to be performed ASAP after filtering). `FeedLoadable` has accordingly been updated to conform to `Filterable`.

The filters can be tested in action using the `eric/new-filtering` branch on `Mlem`; in the tile post layout, there is an obnoxious "hide/show read" button on the first entry. I intend to move this to the nav menu shortly. That branch also currently sets the feed loader to `.topYear`, which is a very stable feed so hide/show read behavior can be easily verified.


Sources/MlemMiddleware/FeedLoaders/Filtering/Filters/Post/PostDedupeFilter.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  PostDedupeFilter.swift
+//
+//
+//  Created by Eric Andrews on 2024-05-31.
+//
+
+import Foundation
+
+class PostDedupeFilter: FilterProviding {
```
Sjmarf: I wonder if it would be possible for this to handle other types too? At some point we'll either need an equivalent filter for comments, or it might be possible to have this filter handle both types. Maybe it could be generic, with `FilterTarget` constrained to `ActorIdentifiable`? Or alternatively it could be non-generic and `FilterTarget` could be _equal_ to `ActorIdentifiable` 🤔

Sources/MlemMiddleware/FeedLoaders/Filtering/Filters/Post/PostKeywordFilter.swift
```diff
@@ -0,0 +1,32 @@
...
+class PostKeywordFilter: FilterProviding {
+    typealias FilterTarget = Post2
+    
+    var numFiltered: Int = 0
+    private var keywords: [String]
+    var active: Bool = true
+    
+    init(keywords: [String]) {
+        self.keywords = keywords
+    }
```
Sjmarf: nit

```suggestion
    private var keywords: Set<String>
    var active: Bool = true
    
    init(keywords: Set<String>) {
        self.keywords = keywords
    }
```

Sources/MlemMiddleware/FeedLoaders/Filtering/Filterable.swift
```diff
@@ -0,0 +1,12 @@
...
+//  Filterable.swift
+//
+//
+//  Created by Eric Andrews on 2024-06-03.
+//
+
+import Foundation
+
+public protocol Filterable {
+    associatedtype Filters
```
Sjmarf: I don't think the name "Filters" should be plural. Maybe `FilterType` would be a better name?

Sources/MlemMiddleware/FeedLoaders/Filtering/PostFilter.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  PostFilter.swift
+//
+//
+//  Created by Eric Andrews on 2024-05-31.
+//
+
+import Foundation
+
+public enum PostFilters {
```
Sjmarf: Similar to the comment above; maybe this would be better named `PostFilterType`? That would match our `PostType` enum better.

Sjmarf (changes requested):
Looks great! A few small comments :)

Sources/MlemMiddleware/FeedLoaders/Filtering/Filters/Post/PostDedupeFilter.swift
EricBAndrews: Non-generic doesn't work because of `MultiFilter` conformance requirements, but generic works a charm

Sjmarf (approved):
LGTM 👍

PR #19, opened 2024-06-12
Upgrade system changes

Sjmarf:
Changes to support [my Mlem PR](https://github.com/mlemgroup/mlem/pull/1087). The main change here is that `upgrade()` now only goes up one level at a time. This means that higher-tier models appear in `ContentLoader` as soon as they're available. Previously, nothing would be shown until the upgrade was fully completed.


Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -215,6 +215,8 @@ extension ApiClient: CacheIdentifiable {
 
 extension ApiClient: ActorIdentifiable {
     public var actorId: URL { baseUrl }
+    
+    public var host: String? { actorId.host() }
```
EricBAndrews: We use this exact line a couple other places, probably worth defining it as an extension of `ActorIdentifiable`

Sources/MlemMiddleware/Content Models/FederationStatus.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  File.swift
```
EricBAndrews:
```suggestion
//  FederationStatus.swift
```

EricBAndrews (changes requested):
Code looks good, though I'm not convinced the tier-by-tier upgrade is necessary. The user still needs to wait one full API call before meaningful content is rendered; is it really that much faster to perform a one-tier upgrade than to hop straight to the fully upgraded model?

I don't have any quantitative data, but intuitively it doesn't seem like the full upgrade should be that much slower. If that's the case, then I'm reluctant to introduce this increase in complexity and API usage for a marginal speedup to the first meaningful render. As I said, though, I don't have numbers on this--I'd be easily convinced by a concrete example of where this enhances the user experience.

Sjmarf:
> Code looks good, though I'm not convinced the tier-by-tier upgrade is necessary. The user still needs to wait one full API call before meaningful content is rendered; is it really that much faster to perform a one-tier upgrade than to hop straight to the fully upgraded model?

It makes the profile part of `PersonView` load faster, though loading the posts and comments takes the same amount of time. Here's a side-by-side (this PR is on the left):

https://github.com/mlemgroup/MlemMiddleware/assets/78750526/6a8acf30-08b0-4501-8cef-da95084a505a


I agree this is only a slight advantage. The other reason why it's done this way is to handle the fact that `GetPersonDetailsRequest` is used to fetch pages of the profile. In the corresponding Mlem PR, this is handled by overriding the Person2 -> Person3 upgrade and extracting the posts that way:

```swift
ContentLoader(model: person) { person in
    // ...
} upgradeOperation: { model, api in
    try await model.upgrade(api: api) { entity in
        if let entity = entity as? any Person1Providing {
            let response = try await entity.getPosts(page: 1, limit: 3)
            Task { @MainActor in
                posts = response.posts
            }
            return response.person
        }
        return try await entity.upgrade()
    }
}
```

As this requires overriding a specific stage of the upgrade cycle, I thought this would be easier to do with the multi-stage upgrade system. An easy alternative would be to make a duplicate request, but that's not ideal.

There's also the hypothetical situation where we'd have a `ContentLoader` that needs to load a `Person2` but a `Person3` isn't required. This would be easy to program with a multi-stage upgrade system - if everything is done at once, we'd be making unnecessary requests.

EricBAndrews (approved):
Wow, that's much more significant than I expected--definitely worth the effort.

PR #20, opened 2024-06-15
Thread Safe Caching

### LINKED ISSUE #16: Use actors for content caches
I just got an `EXC_BAD_ACCESS` on the cache dictionary insertion. We probably need to use actors here (at least, for insertion).

EricBAndrews:
This PR reworks caching to be thread-safe.

There is an accompanying PR on the frontend: https://github.com/mlemgroup/mlem/pull/1094

The critical change is in `CoreCache`: rather than using a simple dictionary, `CoreCache` now uses an `ItemCache`, which in turn wraps an `Atomic` dictionary. This ensures that no two threads try to access the dictionary at once, while still providing synchronous cache access.

The `Atomic` is taken from here: https://github.com/ReactiveCocoa/ReactiveSwift; its license, along with all other licenses for packages used, has been added to the new NOTICE.md.

I have opted for the `Atomic` over `actor` because `actor` forces every single cache access to be `async`, which is, simply put, awful. You can witness the atrocities yourself [here](https://github.com/mlemgroup/mlem/compare/00e909caf2cdf20bccbf8989fc2fb244169b6f59...3f3f4921031d19e9f9d63bc49f36027a66a13fc2).

Closes #16.


Sources/MlemMiddleware/API Client/Caching/Atomic.swift
```diff
@@ -0,0 +1,284 @@
...
+
+import Foundation
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+import MachO
+#endif
+
+/// A simple, generic lock-free finite state machine.
+///
+/// - warning: `deinitialize` must be called to dispose of the consumed memory.
+internal struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue == Int32 {
```
Sjmarf:
```suggestion
private struct UnsafeAtomicState<State: RawRepresentable> where State.RawValue == Int32 {
```

Sources/MlemMiddleware/API Client/Caching/Atomic.swift
```diff
@@ -0,0 +1,284 @@
...
+            }
+            return false
+        }
+    }
+#endif
+}
+
+/// `Lock` exposes `os_unfair_lock` on supported platforms, with pthread mutex as the
+/// fallback.
+internal class Lock {
```
Sjmarf:
```suggestion
private class Lock {
```

Sjmarf (approved):
Nice; great job! A couple of nits regarding access level

PR #22, opened 2024-06-22
Update with `MlemApiTypes` changes

Sjmarf:
- Updated the MlemApiTypes submodule to include [these changes](https://github.com/mlemgroup/MlemApiTypes/pull/12).
- Removed `GetPostsRequest` in favor of the auto-generated version
- Added `@Observable` to `ApiClient` (this is needed so that a view update is sent when an ApiClient's permissions change)


EricBAndrews (approved):
Nice! 👍

PR #25, opened 2024-07-02
Blocking, Searching

### LINKED ISSUE #5: Remove ContentModelIdentier
`ContentModelIdentifier` is redundant with `actorId`, should be removed.

Sjmarf:
- Added search and block methods.
- Added `SelectableContentProviding` for entities that can have the "Select Text" action.
- Removed `uid` (closes #5).
- Added `refresh` methods to `Any...` models.
- Renamed `updateSave`, `updateBlock`, `toggleSave` etc to `updateSaved`, `updateBlocked`, `toggleSaved` etc (lmk if you disagree with this)


Sources/MlemMiddleware/Content Models/Comment/AnyComment.swift
```diff
@@ -67,4 +67,13 @@ public extension AnyComment {
             upgradeOperation: upgradeOperation ?? { try await $0.upgrade() }
         )
     }
+    
+    func refresh(upgradeOperation: ((any Base) async throws -> any Base)?) async throws {
```
EricBAndrews, line 71: Maybe nicer to make this a default and avoid the unwrap below?

```suggestion
    func refresh(upgradeOperation: ((any Base) async throws -> any Base) = { try await $0.upgrade() }) async throws {
```

Sources/MlemMiddleware/Content Models/Community/AnyCommunity.swift
```diff
@@ -67,4 +67,13 @@ public extension AnyCommunity {
             upgradeOperation: upgradeOperation ?? { try await $0.upgrade() }
         )
     }
+    
+    func refresh(upgradeOperation: ((any Base) async throws -> any Base)?) async throws {
```
EricBAndrews, line 71: Ditto above

Sources/MlemMiddleware/Content Models/Person/AnyPerson.swift
```diff
@@ -67,4 +67,13 @@ public extension AnyPerson {
             upgradeOperation: upgradeOperation ?? { try await $0.upgrade() }
         )
     }
+    
+    func refresh(upgradeOperation: ((any Base) async throws -> any Base)?) async throws {
```
EricBAndrews, line 71: Ditto above

Sources/MlemMiddleware/Content Models/Post/AnyPost.swift
```diff
@@ -67,4 +67,13 @@ public extension AnyPost {
             upgradeOperation: upgradeOperation ?? { try await $0.upgrade() }
         )
     }
+    
+    func refresh(upgradeOperation: ((any Base) async throws -> any Base)?) async throws {
```
EricBAndrews, line 71: "

Sources/MlemMiddleware/Content Models/Post/Post1Providing.swift
```diff
@@ -77,6 +77,17 @@ public extension Post1Providing {
     }
 }
 
+// SelectableContentProviding conformance
+public extension Post1Providing {
+    var selectableContent: String? {
+        if let content {
+            "\(title)\n\n\(content)"
```
EricBAndrews, line 84: Since we're selecting Markdown, it might be nice to format the title--that should also let us get rid of one newline and kick the title/body content separation down to whatever Markdown handler the selected text gets fed to

```suggestion
            "#\(title)\n\(content)"
```

EricBAndrews (changes requested):
LGTM 💯 

Fare thee well `ContentModelIdentifier` 🫡

Sources/MlemMiddleware/Content Models/Post/Post1Providing.swift
Sjmarf: At the moment we don't render this as markdown in the "Select Text" sheet - we render it as plain text. I *could* render it using the `MarkdownText` view, which would maintain select-ability, if you think that would be better? One of the arguments for keeping it plain-text is that you can copy the markdown delimiters too, which can be useful for quoting someone in a reply and keeping all of the markdown.

EricBAndrews: Ah, my comment was unclear--I agree that we should keep it plain text with Markdown formatting symbols explicitly included. The title is a weird element, since it's rendered separately everywhere but here, where we mash it into the body. We actually have a few things we can do here:
- Keep as-is, leave the title unformatted. This is the simplest, but if the text is pasted verbatim somewhere else, it won't necessarily be clear that the first line is a title.
- Add a header formatting symbol to the title. This will clearly identify the title if the user pastes it somewhere else that consumes Markdown, and is the most Markdownish way to keep title and body semantically distinct.
- Make the title a Markdown link to the post? Maybe useful for keeping track of the source of the content, but could be annoying to remove if the user doesn't want it.

Side note, do we maybe want a "Copy Body" button as well? Should be easy to implement, and I can see it being useful.

Sources/MlemMiddleware/Content Models/Comment/AnyComment.swift
Sjmarf: `Upgradable` requires implementation of the `refresh` method. If it were non-optional and we were calling `refresh` on `any Upgradable`, we'd have to write out the default value every time:

```swift
model.refresh { try await $0.upgrade() }

// Another example, where upgradeOperation is optional
model.refresh(upgradeOperation: upgradeOperation ?? { try await $0.upgrade() })
```

Having it accept `nil` makes the syntax a little simpler:

```swift
model.refresh(upgradeOperation: nil)

// Where upgradeOperation is optional
model.refresh(upgradeOperation: upgradeOperation)
```

Another benefit of having it accept `nil` is that if we ever need to change the default implementation, we could change it easily in one place without having to modify everywhere `{ try await $0.upgrade() }` is written.

Sources/MlemMiddleware/Content Models/Post/Post1Providing.swift
Sjmarf: Hmm. I suppose we could render two individually highlightable text views in a `VStack`, each with a rounded rect background so that they're visually separated. Then there could be a "Copy" icon next to each section to copy the text blocks individually, and the "Copy all" button at the top could copy both?

I started working on an markdown renderer that shows delimiters a while ago, for use in the post/comment composer. That renderer could work well here - it allows the user to both see and copy the markdown. That renderer isn't done yet though, and probably won't be in-time for 2.0. It's not the rendering logic that's the problem - it's being performant enough on large bodies of text. The article you posted about Paper's solution details some of the ways that it could be solved, but it gets quite complex and there are a lot of edge cases to cover.

The implementation we've got right now isn't ideal, but it's identical to 1.0, which is the minimum we need to get 2.0 going. In the interest of getting 2.0 up-to-speed as fast as possible I think we should leave extra improvements out for now, and we can work on making a better version later.

EricBAndrews: Works for me 👍

Sources/MlemMiddleware/Content Models/Comment/AnyComment.swift
EricBAndrews: Fair enough 👍

EricBAndrews (approved):
LGTM 👍

PR #26, opened 2024-07-05
More Feed Types

EricBAndrews:
This PR accompanies Mlem [#1131](https://github.com/mlemgroup/mlem/pull/1131#issue-2392920244)

It makes the following changes:
- Prevent page from being passed to API if cursor provided in GetPosts (otherwise page loading throws a 400 error on pages >= 2)
- Added a `URLRequest` debug print extension that prints out the request in human-readable form
- Added a `FeedLoading` protocol to which `CoreFeedLoader` (and thus all inheriting types) conform
- Fixed a bug where `changeFeedType` wouldn't trigger a reload when feed type changed
- Fixed a bug where `GetPosts` would not respect the `feed` parameter


Sjmarf (approved):
Looks good 💯

PR #27, opened 2024-07-06
Inbox

Sjmarf:
Sorry for the monster PR :/


Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -220,6 +221,16 @@ extension ApiClient: ActorIdentifiable {
     public var actorId: URL { baseUrl }
 }
 
+extension ApiClient: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(self.baseUrl)
+    }
```
EricBAndrews: This looks like it'll hash every `ApiClient` for a given instance to the same value--is this intentional behavior, or do we want to incorporate the user as well to guarantee unique clients?

Sources/MlemMiddleware/Content Models/ContentModel.swift
```diff
@@ -0,0 +1,43 @@
...
+    static var tierNumber: Int { get }
+    var api: ApiClient { get }
+}
+
+public extension ContentModel where Self: ActorIdentifiable {
+    var apiIsLocal: Bool { api.host == host }
+}
+
+public protocol ContentIdentifiable: AnyObject, ContentModel, Hashable, Identifiable {
+    static var modelTypeId: String { get }
```
EricBAndrews: This might be a little nicer as an `enum: String`, since it's a bounded list of unique values that we want to hash

EricBAndrews (changes requested):
Looks good! I like the choice to consolidate mentions and replies into a single model.

A couple minor implementation questions.

Sources/MlemMiddleware/API Client/ApiClient.swift
Sjmarf: Nice catch, I missed that 👍

Sources/MlemMiddleware/Content Models/ContentModel.swift
Sjmarf: Sure. It really is just `ContentModelIdentifier` all over again lol

EricBAndrews (approved):
LGTM! 🔥

PR #28, opened 2024-07-07
Post Feed Loader Cleanup

EricBAndrews:
This PR decouples community and aggregate post feed loading:

- `StandardPostFeedLoader` is now `CorePostFeedLoader`; it provides operations generic to any pure post feed loading loader, requiring the inheriting class to simply override `getPosts`
- Added `AggregatePostFeedLoader` and `CommunityPostFeedLoader`, which handle those respective feed loading operations; these are extremely lightweight subclasses of `CorePostFeedLoader`.
- Moved `preloadImages` out to a standalone function, since we're going to want it for user content feed loading as well

The only loss of functionality is that you now can't swap a single loader from aggregate to community loading, but we never did that anyway.

This requires practically no changes to the frontend, just some name swaps: https://github.com/mlemgroup/mlem/pull/1135


Sources/MlemMiddleware/API Client/Helpers/PreloadImages.swift
```diff
@@ -0,0 +1,55 @@
...
+//  PreloadImages.swift
+//
+//
+//  Created by Eric Andrews on 2024-07-07.
+//
+
+import Foundation
+import Nuke
+
+func preloadImages(
```
Sjmarf: Could this be a method of `Post2Providing`?

Sources/MlemMiddleware/FeedLoaders/Post Feeed Loaders/CorePostFeedLoader.swift
```diff
@@ -0,0 +1,117 @@
...
+import Observation
+
+/// Post tracker for use with single feeds. Can easily be extended to load any pure post feed by creating an inheriting class that overrides getPosts().
+@Observable
+public class CorePostFeedLoader: StandardFeedLoader<Post2> {
+    private(set) var postSortType: ApiSortType
+    
+    // true when the items in the tracker are stale and should not be displayed
+    var isStale: Bool = false
+
```
Sjmarf: Unused?

```suggestion
```

Sjmarf (changes requested):
Looks good!

Sources/MlemMiddleware/API Client/Helpers/PreloadImages.swift
EricBAndrews: It could, but it feels a little weird to put it there since it's a bulk operation. There is a bunch of `Post2`-specific logic in here, though--IMO it would make more sense to keep this function freestanding but have it just take in `[ImageRequest]` instead of `[Post2]` and put the `ImageRequest` generation code in `Post2Providing`

Sjmarf: Ok, that makes sense 👍

I was just looking through the [Nuke docs](https://kean-docs.github.io/nuke/documentation/nuke/prefetching/) and read this: 

> The prefetcher automatically cancels all of the outstanding tasks when deallocated.

Does this mean that the prefetching doesn't actually have time to fetch all of the images when used in a standalone function like this? If so, it might be better to save it as a property of the feed loader.

```swift
// In the feed loader
let requests = posts.flatMap {
    $0.imageRequests(smallAvatarSize: smallAvatarSize, largeAvatarSize: largeAvatarSize)
} 
// Where `prefetcher` is a stored property of the feed loader
prefetcher.startPrefetching(with: requests)
```

EricBAndrews: Ooh, that's a good point. With the image requests kicked down to `Post2Providing` it's also trivial enough to inline this that we don't need the standalone helper.

Sjmarf (approved):
LGTM!

PR #31, opened 2024-07-10
Block List

### LINKED ISSUE #29: `BlockList`
Similar to `SubscriptionList`. Should store the `id` and `actorId` of every blocked person, community and instance. This will be useful for keeping track of block status - it's not always included in certain responses (particularly for instances - the *only* way to get instance's block status is to request the full block list). 

Additionally it would be useful to have this for multi-accounts - we can union the block-lists of each account easily to decide whether to show a certain post/comment.

Sjmarf:
Adds a `BlockList` class, an instance of which is stored in `ApiClient` in a similar fashion to `SubscriptionList`. The `BlockList` stores only the IDs and actor IDs of blocked entities - it does not store full content models like `SubscriptionList` does. There are two main reasons why we need this class:

1) The Lemmy API generally doesn't tell us when an entity is blocked in very many API types. For instances, this information isn't supplied *at all* outside of specifically fetching the instance block list. Keeping the blocklist in memory allows us to keep all tiers of model "blockable" at all times.

2) When we do multi-accounts we'll need to hide any post/comment by an author blocked on *any* account. We can do this easily by iterating over each logged-in session and checking the `BlockList` of each.

Later down the line we could consider adding a `RichBlockList` that *does* store full content models, for use in the "blocklist" page. That might be overkill though - the lack of state-faking across multiple tabs probably won't be noticed as much in that case.

I also added instance block methods in this PR.

Closes #29


EricBAndrews (approved):
LGTM! 🚀

PR #32, opened 2024-07-18
Made `UnreadCount.update` `@MainActor`

EricBAndrews:
Annotates `UnreadCount.update` with `@MainActor`. This fixes a crash I was experiencing when switching to the inbox.


Sjmarf (approved):
Nice 👍

PR #33, opened 2024-07-22
More operations and awaiting operation results

Sjmarf:
Added methods for replying, editing, deleting, reporting and hiding.

Because `update` and `toggle` methods of content models are synchronous, there wasn't a way to check whether an operation succeeded or not from the Mlem code. Being able to do this is important for higher-risk operations such as deleting a post, where we'd want to let the user know if it fails. I've resolved this by making all `update` and `toggle` methods return `Task<StateUpdateResult, Never>`, where `StateUpdateResult` is an enum defined as follows:

```swift
public enum StateUpdateResult {
    case succeeded
    case failed
    /// Returned when the action is queued for later, e.g. when a post is marked as read.
    case deferred
    case ignored
}
```

Callers can await the result of this task if needed like so:

```swift
func deletePost() {
    Task {
        let operation = post.updateDeleted(true)
        if await operation.result.get() == .failed {
            ToastModel.main.add(.failure("Failed to delete post!"))
        }
    }
}
```


Sources/MlemMiddleware/Content Models/Community/Community2Providing.swift
```diff
@@ -64,25 +68,28 @@ public extension Community2Providing {
...
+    @discardableResult
+    func toggleFavorite() -> Task<StateUpdateResult, Never> {
         updateFavorite(!favorited)
     }
     
-    func updateFavorite(_ newValue: Bool) {
+    @discardableResult
+    func updateFavorite(_ newValue: Bool) -> Task<StateUpdateResult, Never> {
         guard let subscriptions = self.api.subscriptions else {
             print("Tried to toggle favorite, but no SubscriptionList found!")
```
EricBAndrews: *nit*

This should maybe be an `assertionFailure`? Not sure when we'd hit this unless something has gone wrong

Sources/MlemMiddleware/Content Models/Post/Post2Providing.swift
```diff
@@ -89,6 +97,19 @@ public extension Post2Providing {
...
     }
     
+    @discardableResult
+    func toggleHidden() -> Task<StateUpdateResult, Never> {
+        updateHidden(!hidden)
+    }
+    
+    @discardableResult
+    func updateHidden(_ newValue: Bool) -> Task<StateUpdateResult, Never> {
+        // Unlike other
```
EricBAndrews: ?

Sources/MlemMiddleware/Content Models/ReportableProviding.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  File.swift
```
EricBAndrews: *nit*
```suggestion
//  ReportableProviding.swift
```

EricBAndrews (approved):
LGTM! 🚀 

Left a couple little comments, nothing blocking

Sources/MlemMiddleware/Content Models/Post/Post2Providing.swift
Sjmarf: Oops, must have started typing it and then forgot to finish it haha

Sources/MlemMiddleware/Content Models/Community/Community2Providing.swift
Sjmarf: Sure. It's technically possible to hit it, but very unlikely, and `assertionFailure` won't cause a crash in the release build anyways

PR #34, opened 2024-07-24
User Content Feed Loader

EricBAndrews:
This PR adds a feed loader for user content (mixed posts and comments, used in saved and profile).

For testing, you can use https://github.com/mlemgroup/mlem/pull/1157, which implements the saved feed.

Due to the fact that the API does not provide a type-specific endpoint to load this content, we can't just use the parent/child loader setup; instead, I have implemented a similar architecture but using a much more lightweight "stream" in place of the child loaders. All of this is completely non-generic because as far as I know this is the only case where we have this API constraint, and for all other mixed feeds we can just use the parent/child architecture.

The basic loading concept works like this:
- `UserContentStream` stores a list of items, their loading thresholds, and their stream cursor. Interaction with them is done through three methods:
  - `addItems`: adds new items to the stream. The stream handles updating the loading thresholds.
  - `nextItemSortVal`: functions identically to the child feed loader; returns the sort value of the next item in the stream, but does not affect the cursor
  - `consumeNextItem`: returns the next item in the stream and updates the cursor
  - The `UserContentStream` also provides the convenience `needsMoreItems` bool, which is true when the stream believes there are more items to load but has reached the end of its items.
- `UserContentFeedLoader` maintains two `UserContentStream`s, one for posts and one for comments
- To load more items, the `UserContentFeedLoader` performs the same stream merge operation as the parent/child architecture. Unlike a parent tracker, however, at each iteration the `UserContentFeedLoader` checks whether *either* child needs more items and, if so, triggers a load.

To support single-type profile feeds, `UserContentFeedLoader` boasts some additional modifications from the standard feed loading architecture:
- `loadIfThreshold` supports checking a particular stream's thresholds. This is not currently used, but is designed to support single-item profile feeds.
- The streams themselves are private, since they are directly and intimately managed by the `UserContentFeedLoader` and have lots of thread-unsafe operations. Single-type feeds are therefore exposed through the `posts` and `comments` computed vars.
- To support refreshing without clearing on single-type feeds, two temporary streams are defined which supply the single-type streams while the main streams are reset. This is required because loading a fresh stream involves first clearing the children, then loading new items into them, then finally merging those items together into the parent stream.

This PR also makes a few minor changes:
- `MiddlewareConstants.infiniteLoadThresholdOffset` is now 10 instead of -10. Arithmetic has been updated accordingly.
- `getContent` now returns a tuple of `(posts: [Post2], comments: [Comment2])` instead of a `Person3`, and takes parameters
- `FeedLoaderSortVal.published` has been renamed to `.new`. This will let us support `.old` in the future.
- `FeedLoader` thresholds are now handled using the new `Thresholds` struct.


Sources/MlemMiddleware/API Client/ApiClient+Person.swift
```diff
@@ -95,24 +95,27 @@ public extension ApiClient {
...
-            limit: 1,
+            sort: sort,
+            page: page,
+            limit: limit,
             communityId: nil,
-            savedOnly: nil
+            savedOnly: savedOnly
         )
         let response = try await perform(request)
-        return caches.person3.getModel(api: self, from: response)
```
Sjmarf, line 115: We should continue to include `Person3` as part of the return type here, because we'll need to call this method from `PersonView` to both upgrade Person2 to Person3 and load the first page of content in a single request (see [here](https://github.com/mlemgroup/MlemMiddleware/pull/19#issuecomment-2169269847) for explanation and example code).

Sources/MlemMiddleware/Content Models/Person/Person1Providing.swift
Sjmarf, line 1: `Person1Providing` defines a `getPosts` method - ideally we'd replace that with `getContent` that returns both posts and comments.

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContent.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+//  Created by Eric Andrews on 2024-07-21.
+//
+
+import Foundation
+
+public class UserContent: Hashable, Equatable, FeedLoadable {
+    public typealias FilterType = UserContentFilterType
+    
+    public let wrappedValue: Value
```
Sjmarf: Would it be possible to use an enum directly here rather than a class with an enum inside of it? Or would that not work?

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
```diff
@@ -0,0 +1,240 @@
...
+        }
+        postStream = .init()
+        commentStream = .init()
+        apiPage = 0
+        contentPage = 0
+        
+        try await loadMoreItems()
+        
+        tempPostStream = nil
+        tempCommentStream = nil
```
Sjmarf: If an error is thrown here the temp streams won't be erased

```suggestion
        defer {
            tempPostStream = nil
            tempCommentStream = nil
        }
        try await loadMoreItems()
```

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
```diff
@@ -0,0 +1,240 @@
...
+/// which is then incorporated into both child streams.
+
+@Observable
+public class UserContentFeedLoader: FeedLoading {
+    public var api: ApiClient
+    public var items: [UserContent]
+    public var loadingState: LoadingState
+    
+    // loading configuration
+    private(set) var sortType: FeedLoaderSort.SortType
```
Sjmarf: Will this need to be read from Mlem?

```suggestion
    public private(set) var sortType: FeedLoaderSort.SortType
```

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
```diff
@@ -0,0 +1,240 @@
...
+/// just posts or just comments, and thus the standard Parent/Child FeedLoader construction does not work without
+/// severe API waste. This solution is a simplified variant of that architecture.
+///
+/// The UserContentFeedLoader is the parent loader. It is responsible for all data fetching, and keeps track of two
+/// UserContentStreams, one for Posts and one for Comments. To load a page of items, it consumes and merges the child streams, just as
+/// in the standard Parent/Child FeedLoader; if either stream reaches the end of its items, it triggers a new load, the response from
+/// which is then incorporated into both child streams.
+
+@Observable
+public class UserContentFeedLoader: FeedLoading {
```
Sjmarf: If we're keeping with the "person" naming scheme over "user", this should be called `PersonContentFeedLoader`

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
```diff
@@ -0,0 +1,240 @@
...
+    
+    // loading configuration
+    private(set) var sortType: FeedLoaderSort.SortType
+    private var userId: Int
+    private var savedOnly: Bool
+    
+    // prefetching
+    private let smallAvatarIconSize: Int
+    private let largeAvatarIconSize: Int
+    private let urlCache: URLCache
```
Sjmarf: Unused?

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
```diff
@@ -0,0 +1,240 @@
...
+    public func loadIfThreshold(_ item: UserContent) throws {
+        try loadIfThreshold(item, asChild: false)
+    }
+    
+    /// Given a UserContent, loads more items if that content is a threshold item
+    /// - Parameters:
+    ///   - item: UserContent
+    ///   - loadChildOnly: if true, the item will be evaluated against the relevant stream threshold rather than the parent threshold
+    public func loadIfThreshold(_ item: UserContent, asChild: Bool) throws {
+        var shouldLoad: Bool = false
```
Sjmarf: _Nit_

```suggestion
        let shouldLoad: Bool
```

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
```diff
@@ -0,0 +1,240 @@
...
+    private func fetchItems() async throws -> (posts: [Post2], comments: [Comment2]) {
+        return try await api.getContent(authorId: userId, sort: .new, page: apiPage, limit: 50, savedOnly: savedOnly)
+    }
+    
+    /// Loads the next page from the API
+    /// - Warning: This is NOT a thread-safe function! It should only be called from a concurrency-controlled environment
+    private func loadNextApiPage() async throws {
+        apiPage += 1
+        let response = try await fetchItems()
+        preloadImages(response.posts) // TODO: comment images?
```
Sjmarf: > `// TODO: comment images?`

I'll be adding a way to extract images from a markdown tree at some point (issue [here](https://github.com/mlemgroup/LemmyMarkdownUI/issues/17)).

Sjmarf (changes requested):
Nice! This is a clean solution 💯

Sjmarf:
> as far as I know this is the only case where we have this API constraint

There's also the Modlog

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContent.swift
EricBAndrews: Not as it currently stands, since `FeedLoadable` is `AnyObject`, which `enum` cannot conform to.

I'm not actually sure exactly why that conformance is in place, but until we have all the `FeedLoader`s implemented I don't want to modify it because *presumably* I put it there for a reason. I've opened #35 to track this.

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
EricBAndrews: Nice, I've opened #36 to circle back to this when that's done

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentFeedLoader.swift
EricBAndrews: Nice catch!

EricBAndrews:
> > as far as I know this is the only case where we have this API constraint
> 
> There's also the Modlog

Modlog is all a single endpoint, but you can specify the specific type to fetch so we can paginate independently--this is the only endpoint that forces you to get *everything*

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContent.swift
Sjmarf, line 1: This file should be renamed to `PersonContent`

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentProviding.swift
Sjmarf, line 1: ditto

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/UserContentStream.swift
Sjmarf, line 1: ditto

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/PersonContentFeedLoader.swift
```diff
@@ -13,26 +13,25 @@ import Nuke
...
-/// The UserContentFeedLoader is the parent loader. It is responsible for all data fetching, and keeps track of two
-/// UserContentStreams, one for Posts and one for Comments. To load a page of items, it consumes and merges the child streams, just as
+/// The PersonContentFeedLoader is the parent loader. It is responsible for all data fetching, and keeps track of two
+/// PersonContentStreams, one for Posts and one for Comments. To load a page of items, it consumes and merges the child streams, just as
 /// in the standard Parent/Child FeedLoader; if either stream reaches the end of its items, it triggers a new load, the response from
 /// which is then incorporated into both child streams.
 
 @Observable
-public class UserContentFeedLoader: FeedLoading {
+public class PersonContentFeedLoader: FeedLoading {
```
Sjmarf, line 22: WRT [this comment](https://github.com/mlemgroup/MlemMiddleware/pull/34#discussion_r1690536416), we'd need a way to supply the first page of results to `PersonContentFeedLoader` from Mlem. Is there a way to do this? 

This could be done in the initialiser or as an argument of `refresh` of something like that? Or alternatively `refresh` could perform the request and return the `Person3` so that Mlem can use it.

Sjmarf (changes requested):

Sources/MlemMiddleware/FeedLoaders/User Content Feed Loader/PersonContentFeedLoader.swift
EricBAndrews: I've added an optional `withContent` parameter that pre-populates the streams with the given content. This is a tiny bit unsafe since it assumes that the provided content is precisely the first page--the `refresh` idea would be a little safer but add a chunk of complexity, since `loadMoreItems` isn't actually guaranteed to make an API call. Since we standardize all our pagination with `InternetSpeed.pageSize`, I think the cleaner code this approach offers outweighs the risk.

Sjmarf: Nice 👍

Sjmarf (approved):
LGTM!

PR #39, opened 2024-07-31
Website Complex

EricBAndrews:
Minor changes to support the website complex:
- `.link` post type now has the new `PostLink` associated value, which bundles the information needed to display a meaningful link preview
- `PostEmbed` members are now public

Unrelated, but it fixes the `Equatable` conformance on `VotesModel`--previously the `==` method did not return any value.


Sjmarf (approved):
LGTM 👍

PR #42, opened 2024-08-04
Avoid unnecessary view updates

### LINKED ISSUE #1185: [2.0] Lag whilst comments are loading in and when refreshing

Sjmarf:
See also [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/13).

When a property of an `Observable` model was set, views were updated even if the new value was equal to the existing value. When `getComments` was called, `update` will be called on the parent post many times because each `Comment2` contains a `Post1`. This caused the seen view lag. I've fixed this by adding `!=` checks when updating property values.

Also added a few missing properties to `Post`

Closes https://github.com/mlemgroup/mlem/issues/1185


EricBAndrews (approved):
Nice! That's noticeably faster 🚀

PR #44, opened 2024-08-10
Community subscription state-faking

Sjmarf:
The community subscription count is now state-faked together with the subscription status in a `SubscriptionModel`. This is similar to how vote status is combined with vote count in `VotesModel`. This means that when you subscribe to a community, you can see the count change immediately.

Also made some tweaks to the feed loader


EricBAndrews (approved):
LGTM!

PR #46, opened 2024-08-18
Fix FeedLoader early stop

EricBAndrews:
Fixes a cluster of related cursor bugs:
- Fixes a bug in the cursor logic where the StandardFeedLoader would detect EOF prematurely. This was happening due to `cursor` being used for all fetches inside `loadCursorHelper` but not being updated.
- Fixes `fetchPageHelper` using page-based pagination for fetches after page 1 on instances which support cursor-based pagination. On such instances, page-based fetching is only used for the first page.
- Moves EOF cursor detection logic into `FetchResponse.hasContent`


Sjmarf (approved):
LGTM 👍

PR #47, opened 2024-08-19
Person Content

EricBAndrews:
Minor changes to support profile:
- Made `PersonContentStream` use `PersonContent` instead of `Item`. This turns out to be far more ergonomic on the frontend. Type purity within a stream is still guaranteed by restricting modification of `items` to `init` and `addItems`, both of which take `[Item]` and convert internally to `PersonContent`
- `loadingState` modification is now `@MainActor`


Sjmarf (approved):
Nice! 💯

PR #49, opened 2024-08-21
Mark Inbox item as read on reply creation

Sjmarf:
Also added a `cachedValueAcceptable` parameter to `Comment2.getParent`, which I need in the "Edit Comments" PR.


EricBAndrews (approved):
LGTM!

PR #51, opened 2024-08-22
`UnreadCount` reliability

Sjmarf:
Fixed a bug where the inbox refresh popup was incorrectly shown if you happened to mark an item as read at the same time as the count was refreshing.

Non-breaking


EricBAndrews (approved):
LGTM!

PR #53, opened 2024-08-28
Eric/mark read on scroll

EricBAndrews:
Some minor changes to support mark read on scroll:

- Swapped the default behavior of `Post2.updateRead` to not flush the queue--this prevents queued posts from being marked read when a post is interacted with (side note: we don't actually use that toggle anywhere--maybe we can remove it and add it back if we need it?). Flushing the queue is now handled on the UI side.
- Added convenience value `batchMarkReadEnabled` to `ApiClient`

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
@@ -107,18 +107,17 @@ public extension ApiClient {
     }
     
     /// Mark the given post as read. Works on all versions.
-    /// On v0.19.0 and above, calling this will also mark any queued posts as read unless `includeQueuedPosts` is set to `false`.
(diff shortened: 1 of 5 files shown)


Sjmarf:
> this prevents queued posts from being marked read when a post is interacted with (side note: we don't actually use that toggle anywhere--maybe we can remove it and add it back if we need it?).

The idea was that if we're sending a single mark-read request, we may as well flush the queue in the same request so that the server receives the updated read statuses as quickly as possible. In theory it slightly reduces the number of requests we need to send to Lemmy - if a user opens a post to read the comments, any timer-based flushes that occur whilst the user is reading the post won't need to be sent because the queue has already been emptied when the user opened the post.

IMO we should keep it enabled, but it doesn't really matter either way.

Sjmarf (approved):

PR #55, opened 2024-09-02
0.18x Compatibility

EricBAndrews:
Adds logic to include the "auth" body param when submitting requests to instances running < 0.19.

The `eric/beehaw-compatibility` branch on Mlem is hooked up to this branch for testing.

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -160,18 +160,27 @@ public class ApiClient {
     }
     
     private func execute(_ urlRequest: URLRequest) async throws -> (Data, URLResponse) {
+        var urlRequest: URLRequest = urlRequest // make mutable
         print("DEBUG executing...")
         
-        let version = try await version
-        
-        print("DEBUG got version \(version  )")
-        
         // add 0.18x "auth" param if on 18.x instance and token defined
-        if version < .v19_0, let token, let httpBody = urlRequest.httpBody {
-            var body = JSON(httpBody)
-            print("DEBUG \(body)")
-            body = try body.merged(with: JSON.init(dictionaryLiteral: ("auth", token)))
-            print("DEBUG post-merge \(body)")
+        // need to use fetched
+        // if version < .v19_0, let token, let httpBody = urlRequest.httpBody {
+        
+        if fetchedVersion ?? .v18_0 < .v19_0, let token {
+            let authBody: JSON = .init(dictionaryLiteral: ("auth", token))
+            let newBody: JSON
+            if let httpBody = urlRequest.httpBody {
+                print("DEBUG adding to \(httpBody)")
+                newBody = try JSON(httpBody).merged(with: authBody)
+            } else {
+                print("DEBUG creating new body")
+                newBody = authBody
+            }
+            
+            print("DEBUG post-merge \(newBody)")
+            
+            urlRequest.httpBody = try newBody.rawData()
         } else {
             print("DEBUG will not add to \(urlRequest.description)")
         }
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -167,7 +167,16 @@ public class ApiClient {
         // need to use fetched
         // if version < .v19_0, let token, let httpBody = urlRequest.httpBody {
         
-        if urlRequest.httpMethod != "GET", fetchedVersion ?? .v18_0 < .v19_0, let token {
+        let apiVersionRecognized: Bool
+        if case .other = fetchedVersion {
+            apiVersionRecognized = false
+        } else {
+            apiVersionRecognized = true
+        }
+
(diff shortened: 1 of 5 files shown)


Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -159,6 +160,30 @@ public class ApiClient {
...
     private func execute(_ urlRequest: URLRequest) async throws -> (Data, URLResponse) {
+        var urlRequest: URLRequest = urlRequest // make mutable
+        
+        // add 0.18x "auth" param if on 18.x (or unrecognized) instance
+        let apiVersionRecognized: Bool
+        if case .other = fetchedVersion {
+            apiVersionRecognized = false
+        } else {
+            apiVersionRecognized = true
+        }
```
Sjmarf: *nit*
Could this logic be stored as a computed property of `SiteVersion` maybe? 

```swift
var requiresAuthFieldInRequestBody: Bool {
   self == .other || self < .v19_0
}
```

Then the `if` statement on line 174 would be:

```swift
if let token, fetchedVersion?.requiresAuthFieldInRequestBody ?? true, urlRequest.httpMethod != "GET" {
```

(Tbh we should probably have quite a few computed properties on `SiteVersion` for various features - `allowsBatchImageUploads` could go there too rather than on `ApiClient`. OOS though)

Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -130,7 +131,7 @@ public class ApiClient {
     func perform<Request: ApiRequest>(_ request: Request) async throws -> Request.Response {
         let urlRequest = try urlRequest(from: request)
         // this line intentionally left commented for convenient future debugging
-        // urlRequest.debug()
+        urlRequest.debug()
```
Sjmarf:
```suggestion
        // urlRequest.debug()
```

Sjmarf (approved):
Nice! This is a cleaner solution. One minor nit; non-blocking

Sources/MlemMiddleware/Content Models/Internal/SiteVersion.swift
```diff
@@ -50,6 +50,27 @@ public enum SiteVersion: Equatable, Hashable {
...
+        }
+    }
+    
+    /// Checks whether this SiteVersion supports the given feature. Always returns false if version unknown.
+    func suppports(_ feature: Feature) -> Bool {
+        switch self {
+        case .other: false
+        default: feature.minimumVersion <= self
+        }
+    }
```
Sjmarf, line 71: `public`?

Sources/MlemMiddleware/Content Models/Internal/SiteVersion.swift
```diff
@@ -50,6 +50,27 @@ public enum SiteVersion: Equatable, Hashable {
...
+            case .headerAuthentication: .v19_0
+            case .batchMarkRead: .v19_0
+            }
+        }
+    }
+    
+    /// Checks whether this SiteVersion supports the given feature. Always returns false if version unknown.
+    func suppports(_ feature: Feature) -> Bool {
+        switch self {
+        case .other: false
```
Sjmarf, line 68: Should we assume the feature is supported if `.other`?

```suggestion
        case .other: true
```

Sources/MlemMiddleware/Content Models/Internal/SiteVersion.swift
EricBAndrews: Implicitly `public` because the whole extension is

Sources/MlemMiddleware/Content Models/Internal/SiteVersion.swift
EricBAndrews: `false` is the safer failure state:

| | Actually True | Actually False |
| --- | --- | --- |
| Assumed True | ✅ | Worst-case: failed  API calls |
| Assumed False | Worst-case: don't support all possible functionality | ✅ |

Sources/MlemMiddleware/Content Models/Internal/SiteVersion.swift
Sjmarf: Ah, I missed that it was in an extension 👍

Sources/MlemMiddleware/Content Models/Internal/SiteVersion.swift
Sjmarf: That's a good point, for this particular case. Though if we plan to add cases of `Feature` such as `supportsPostHiding` etc, defaulting to `true` would be better for those cases:

| | Actually True | Actually False |
| --- | --- | --- |
| Assumed True | ✅ | Worst-case: "error" toast shown when user attempts to use feature |
| Assumed False |  Worst-case: User can't use feature at all even though Lemmy supports it  | ✅ |

Potentially we could decide whether to default to `true` or `false` via an `isNecessary: Bool` computed property of `Feature` or something. Happy to leave OOS though

PR #66, opened 2024-10-10
Remove Content

Sjmarf:


EricBAndrews (changes requested):
Code looks good--what's with all the frontend packages in `Package.resolved`? Build artifact from local development?

Sjmarf:
Yeah, that's strange 🤔 Is that file even meant to be there? If I delete it, Xcode won't create a new one during a build like it does for Mlem. Maybe I should just delete it?

EricBAndrews (approved):
LGTM! 🚀

PR #70, opened 2024-11-08
Actor Feed Loaders

EricBAndrews:
This PR completely reworks the `FeedLoader` architecture to use native Swift `actor`s for concurrency management. This is achieved by the addition of two new components: the `LoadingActor` and the `Fetcher`.

## `LoadingActor`

The `LoadingActor` is an `actor` that is responsible for loading. Because `actor`s are extremely inflexible, this is designed to be as minimal and generic as possible; it can be thought of as the "kernel" of the loading system. The `LoadingActor` defines two operations:
- `load`: loads the next page of items
- `reset`: resets the loading state

The behavior of these operations is subject to the following behavior requirements:
1. It should be possible to call both `reset` and `load` while an existing `load` call is awaiting the server response
2. If `reset` is called while the actor is loading, the existing load should be immediately cancelled
3. If `load` is called while the actor is loading, the new load call should be ignored

These requirements ensure that the frontend is responsive (e.g., refreshing the feed does not require waiting for an existing load to complete just to be thrown out) and duplicate loads (e.g., as caused by scrolling fast and hitting both the normal and the fallback threshold) are ignored.

To support this behavior, the `LoadingActor` stores a `loadingTask`. This is an independent `Task` that performs the actual server call; if no server call is in progress, `loadingTask` is nil. `load` spawns this task then awaits its result; the `await` operation frees up the actor to handle new `reset` or `load` calls. This architecture satisfies the above requirements:
1. New `load` or `reset` operations can be handled as soon as the `loadingTask` is spawned.
2. If a `reset` call comes in while `loadingTask` is non-nil, it simply calls `Task.cancel()` on the `loadingTask`. That `loadingTask` will return a `cancelled` status.
3. If a `load` call comes in while `loadingTask` is non-nil, `load` returns the `ignored` status.

## `Fetcher`

Swift does not allow inheritance with `actor`s. This renders `LoadingActor` highly inflexible; each fetch use cases requires tracking a unique set of fetch parameters, which is inelegant to handle within the `LoadingActor` itself. `LoadingActor` is therefore instantiated with a `Fetcher`: a class which provides the `fetchPage` and `fetchCursor` methods and handles the unique fetch parameters for each loader. The `Fetcher` is stored within the `FeedLoader` and passed as a reference into the `LoadingActor`, allowing the `FeedLoader` to modify fetching behavior (e.g., change the query string for a search fetch).

## Responsibility Changes

The `FeedLoader` looks largely the same as before, but with much of its operation now delegated to the `Fetcher` and the `LoadingActor`. Most notably, the `FeedLoader`'s `loadingState` no longer has concurrency management responsibilities. The responsibilities of each component are as follows:

`FeedLoader` | `Fetcher` | `LoadingActor`
---|---|---
Filtering, custom post-fetch processing (e.g., image prefetching), and handling frontend requests | Tracking fetch parameters and defining fetching behavior | Concurrency management

### DIFF
--- a/Sources/MlemMiddleware/FeedLoaders/Community/CommunityFeedLoader.swift
+++ b/Sources/MlemMiddleware/FeedLoaders/Community/CommunityFeedLoader.swift
@@ -7,9 +7,7 @@
 
 import Foundation
 
-class CommunityFetcher: Fetcher {
-    typealias Item = Community2
-    
+class CommunityFetcher: Fetcher<Community2> {
     let api: ApiClient
     var query: String
     var pageSize: Int
@@ -24,7 +22,7 @@ class CommunityFetcher: Fetcher {
         self.sort = sort
     }
     
-    func fetchPage(_ page: Int) async throws -> FetchResponse<Community2> {
+    override func fetchPage(_ page: Int) async throws -> FetchResponse<Community2> {
         let communities = try await api.searchCommunities(
             query: query,
             page: page,
@@ -40,9 +38,9 @@ class CommunityFetcher: Fetcher {
         )
     }
     
-    func fetchCursor(_ cursor: String) async throws -> FetchResponse<Community2> {
-        fatalError("Unsupported loading operation")
-    }
+//    func fetchCursor(_ cursor: String) async throws -> FetchResponse<Community2> {
+//        fatalError("Unsupported loading operation")
+//    }
 }
 
 @Observable
--- a/Sources/MlemMiddleware/FeedLoaders/Community/CommunityFeedLoader.swift
+++ b/Sources/MlemMiddleware/FeedLoaders/Community/CommunityFeedLoader.swift
@@ -7,6 +7,37 @@
 
 import Foundation
 
+struct CommunityFetchProvider: FetchProviding {
+    typealias Item = Community2
+    
+    let api: ApiClient
+    let query: String
+    let pageSize: Int
+    let listing: ApiListingType
+    let sort: ApiSortType
+    
+    func fetchPage(_ page: Int) async throws -> FetchResponse<Community2> {
+        let communities = try await api.searchCommunities(
+            query: query,
+            page: page,
+            limit: pageSize,
+            filter: listing,
+            sort: sort
+        )
+        
+        return FetchResponse<Community2>.init(
+            items: communities,
+            prevCursor: nil,
+            nextCursor: nil,
+            numFiltered: 0
+        )
+    }
+    
+    func fetchCursor(_ cursor: String) async throws -> FetchResponse<Community2> {
+        fatalError("Unsupported loading operation")
+    }
+}
+
 @Observable
 public class CommunityFeedLoader: StandardFeedLoader<Community2> {
     public var api: ApiClient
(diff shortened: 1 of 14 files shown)


Sources/MlemMiddleware/FeedLoaders/Generics/LoadingActor.swift
```diff
@@ -0,0 +1,166 @@
...
+        }
+        
+        do {
+            guard let loadingTask else {
+                assertionFailure("loadingTask is nil!")
+                throw LoadingError.noTask
+            }
+            let result = try await loadingTask.result.get()
+            
+            if case let .done = result {
```
Sjmarf:
```suggestion
            if case .done = result {
```

Sources/MlemMiddleware/FeedLoaders/Generics/StandardFeedLoader.swift
```diff
@@ -9,253 +9,84 @@ import Foundation
...
         var newItems: [Item] = .init()
-        while newItems.count < pageSize {
-            // use cursor-based fetching if available
-            let fetched: FetchResponse<Item>
-            if let loadingCursor {
-                fetched = try await fetchCursor(cursor: loadingCursor)
-            } else {
-                fetched = try await fetchPage(page: page + 1)
-            }
+        var abort: Bool = false // this is slightly awkward but lets us trigger a loop break from within the switch handlers below
```
Sjmarf: [Labelling the loop](https://www.hackingwithswift.com/example-code/language/how-to-break-out-of-multiple-loop-levels-using-labeled-statements) might be a cleaner way of doing this

Sources/MlemMiddleware/FeedLoaders/Community/CommunityFeedLoader.swift
```diff
@@ -7,52 +7,67 @@
...
-            numFiltered: 0
+            nextCursor: nil
         )
     }
     
-    override public func refresh(clearBeforeRefresh: Bool) async throws {
-        try await super.refresh(clearBeforeRefresh: clearBeforeRefresh)
+//    func fetchCursor(_ cursor: String) async throws -> FetchResponse<Community2> {
+//        fatalError("Unsupported loading operation")
+//    }
```
Sjmarf: Dead code?

Sources/MlemMiddleware/FeedLoaders/Generics/LoadingActor.swift
```diff
@@ -0,0 +1,166 @@
...
+                return .success(response.items)
+            } else {
+                page += 1
+                print("[\(Item.self) Fetcher] loading page \(page)")
+                let response = try await fetchPage(page)
+                
+                // if nothing returned, loading is finished
+                if response.items.isEmpty {
+                    return .done(.init())
+                }
```
Sjmarf: Could we assume that there are no more posts if `response.items < pageSize` rather than `response.items.isEmpty`? Doing so would make some pages load quicker - for a community with `< pageSize` posts, Mlem will currently wait for the second page load to be returned before displaying the first page of results. I don't think performing the second page load is necessary?

(This behavior also happened before this PR so feel free to leave OOS)

Sjmarf (changes requested):
Nice work! 🔥  A few minor nits, and I also found a couple of bugs which I mentioned in the other PR's review.

Sources/MlemMiddleware/FeedLoaders/Generics/StandardFeedLoader.swift
EricBAndrews: Nice, didn't know about that!

Sources/MlemMiddleware/FeedLoaders/Generics/LoadingActor.swift
EricBAndrews: That should work, since this check happens before any filtering--I've implemented and added some detailed logging just in case.

EricBAndrews:
Converting back to draft pending frontend changes

Sjmarf (approved):
LGTM! 💯

Sjmarf (changes requested):
(See the other PR)

Sjmarf (approved):
LGTM!

PR #71, opened 2024-11-09
Fix Person1 block bug

Sjmarf:
Fixed a bug in which `Person1.blocked` was assumed to be `false` too early. Setting it to `nil` causes the `Person1` initializer to check if the `Person1` is in the `BlockList`, which was the intended behavior.


EricBAndrews (approved):
Nice 👍

PR #72, opened 2024-11-16
Ban methods

Sjmarf:
Added ban methods.

 I also previously thought that `ApiPostView.bannedFromCommunity` was a renaming of `ApiPostView.creatorBannedFromCommunity` - this isn't the case, which caused every person in a community to show up as banned if *you* were banned from the community.


EricBAndrews (approved):
LGTM ✅

PR #74, opened 2024-11-28
Eric/inbox feed loader

EricBAndrews:
Implements full inbox feed loading and the parent/child feed loading architecture. The core theory is unchanged from v1: the parent feed loader "fetches" by calling the child loaders to pull items out of a stream and merging the results.

Major divergences from the old approach:
- Merging is now done in the `Fetcher`; this enables much more reuse of the `StandardFeedLoader` logic
- Child loaders can only have one parent. We never used the multi-parent functionality of the old approach, and having only one avoids a lot of nasty recursion cases
- Child loaders now share a type with the parent loaders. This removes a couple minor frontend benefits (i.e., being able to directly use the child type when working with a child loader), but makes the generic code _way_ simpler and bypasses a lot of otherwise super thorny filtering problems

Note that it is possible to trip an assertionFailure and induce some slightly anomalous behavior by mashing the filter button a _lot_. This is due to some really nasty concurrency issues that actors aren't _quite_ up to solving (at least, not cleanly)--I've opened #76 to use a DispatchQueue instead of actors for thread safety, but that's a big lift for very minor gain. Since this doesn't crash the app, is easily fixed on reload, and isn't "normal" behavior, I'm content leaving it unfixed.


Sources/MlemMiddleware/Content Models/ReadableProviding.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  ReadableProviding.swift
+//  MlemMiddleware
+//
+//  Created by Eric Andrews on 2024-12-04.
+//
+
+public protocol ReadableProviding {
```
Sjmarf, line 8: *nit* `InboxItemProviding` should conform to this

Sources/MlemMiddleware/FeedLoaders/Filtering/Filters/Post/PostReadFilter.swift
Sjmarf, line 1: File name needs changing

Sources/MlemMiddleware/FeedLoaders/Generics/MultiFetcher.swift
```diff
@@ -0,0 +1,88 @@
...
+        
+        // find the highest-sorted item from the tops of all sources
+        for source in sources {
+            (sortVal, sourceToConsume) = try await compareNextItem(lhsVal: sortVal, lhsSource: sourceToConsume, rhsSource: source)
+        }
+        
+        if let nextItem = sourceToConsume?.consumeNextItem() {
+            // ensure the item is of the correct type
+            assert(nextItem is Item, "Could not convert item to [\(Item.self)]")
+            return nextItem as! Item // this only fails if there's a serious type issue, which should be caught by the assertion above
```
Sjmarf: Might it be better to use `as?` here to avoid a fatal crash? The function can return `nil` anyways

```suggestion
            return nextItem as? Item
```

Sources/MlemMiddleware/FeedLoaders/Inbox/MentionFeedLoader.swift
Sjmarf, line 1: Unused?

Sources/MlemMiddleware/FeedLoaders/Inbox/MessageFeedLoader.swift
Sjmarf, line 1: Unused?

Sources/MlemMiddleware/FeedLoaders/Inbox/ReplyFeedLoader.swift
Sjmarf, line 1: Unused?

Sources/MlemMiddleware/FeedLoaders/Post Feed Loaders/SearchPostFeedLoader.swift
```diff
@@ -14,7 +14,9 @@ public class SearchPostFetcher: PostFetcher {
     public var listing: ApiListingType
     
     // setters to allow manual overriding of these for search use cases
-    public func setApi(_ api: ApiClient) { self.api = api }
+    public override func changeApi(to newApi: ApiClient) async {
+        await super.changeApi(to: newApi)
+    }
```
Sjmarf, line 20: Unnecessary?

```suggestion
```

Sjmarf (changes requested):
Nice! A few minor nits

Sources/MlemMiddleware/FeedLoaders/Generics/MultiFetcher.swift
EricBAndrews: Nice catch! Turns out the whole cast is unnecessary now that child loaders just use the parent type 😁

Sources/MlemMiddleware/FeedLoaders/Post Feed Loaders/SearchPostFeedLoader.swift
EricBAndrews: This makes `changeApi` public for this loader--it's normally internal because we don't need to do this anywhere else

Sources/MlemMiddleware/FeedLoaders/Inbox/MentionFeedLoader.swift
EricBAndrews: Used in `InboxView` as sources for the `InboxFeedLoader`

Sources/MlemMiddleware/FeedLoaders/Generics/Fetcher.swift
```diff
@@ -29,7 +29,7 @@ enum LoadingResponse<Item: FeedLoadable> {
 }
 
 public class Fetcher<Item: FeedLoadable> {
```
Sjmarf, line 34: Any chance fetchers could be made `@Observable` while we're here? Technically OOS but the sort mode icon in the feed occasionally doesn't update because the sort mode status is stored in `PostFetcher`, which isn't `@Observable`.

Sjmarf (approved):
Nice 👍 I mentioned that I found a few bugs on the other PR, I'm not sure whether those require modifying this code or not.

PR #79, opened 2024-12-14
Use @MainActor for state updates

Sjmarf:
This should hopefully resolve some random crashes. This change doesn't require an Mlem PR.


EricBAndrews (approved):
LGTM! 💯

PR #80, opened 2024-12-15
Remove & Purge Community

Sjmarf:
Added remove and purge methods for communities, and moved the `isAdmin` property from `Person4` to `Person2`.

### DIFF
--- a/Sources/MlemMiddleware/Custom API Models/Extensions/APIPersonView+Extensions.swift
+++ b/Sources/MlemMiddleware/Custom API Models/Extensions/APIPersonView+Extensions.swift
@@ -7,4 +7,12 @@
 
 import Foundation
 
-extension ApiPersonView: Person2ApiBacker {}
+extension ApiPersonView: Person2ApiBacker {
+    public var admin: Bool {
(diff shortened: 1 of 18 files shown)


Sources/MlemMiddleware/Custom API Models/Extensions/APIPersonView+Extensions.swift
```diff
@@ -7,4 +7,12 @@
...
 import Foundation
 
-extension ApiPersonView: Person2ApiBacker {}
+extension ApiPersonView: Person2ApiBacker {
+    public var admin: Bool {
+        if let admin = self.isAdmin ?? self.person.admin {
+            return admin
+        }
+        assertionFailure()
+        return false
```
EricBAndrews: *nit*

Perhaps more elegant to solve with a `guard`:

```suggestion
        guard let admin = self.isAdmin ?? self.person.admin else {
            assertionFailure("Could not determine admin status")
            return false
        }
        return admin
```

EricBAndrews (approved):
Nice! 🚀

PR #83, opened 2024-12-17
Report

Sjmarf:


Sources/MlemMiddleware/API Client/ApiClient+Report.swift
```diff
@@ -0,0 +1,83 @@
...
+            page: page,
+            limit: limit,
+            unresolvedOnly: unresolvedOnly
+        )
+        let response = try await perform(request)
+        return await caches.report.getModels(api: self, from: response.privateMessageReports)
+    }
+    
+    @discardableResult
+    func resolveReport(
```
EricBAndrews: Do we want to have this broken out into separate functions for each report type? We define the `Report` abstraction as a wrapper on top of the Lemmy API, so it makes more sense to me for the abstraction work to happen in `Report` and keep `ApiClient` dumb and closely mapped to the Lemmy API.

EricBAndrews (changes requested):
Looks good! One refactoring note.

Sources/MlemMiddleware/API Client/ApiClient+Report.swift
Sjmarf: Sure, that makes sense 👍

EricBAndrews (approved):
LGTM! 🔥

PR #84, opened 2024-12-18
View Votes

Sjmarf:
Added `PersonVote` so that moderators and admins can view a list of votes, and tweaked how community ban status is stored.


EricBAndrews (approved):
Nice! 🚀

PR #86, opened 2024-12-23
Keyword Filters

EricBAndrews:
Adds keyword filtering.

The core architectural tweak here is the addition of `FilterContext`, which is a struct bundling all the information needed to perform filtering (e.g., filters need to know which communities a user moderates). For simplicity of integration with the generic `FeedLoader` structure, this is implemented as a wholly generic struct, so it needs to include all the information needed for any filter; we don't filter much besides posts, however, so I don't expect it to bloat.


Sources/MlemMiddleware/Content Models/Person/Person4Providing.swift
```diff
@@ -89,3 +89,7 @@ public extension Person4Providing {
     var enableAnimatedImages_: Bool? { person4.enableAnimatedImages }
     var collapseBotComments_: Bool? { person4.collapseBotComments }
 }
+
+public extension Person4Providing {
+    var moderatedCommunityIds: Set<URL> { .init(moderatedCommunities.map { $0.actorId }) }
```
Sjmarf: Imo this should be called `moderatedCommunityActorIds` to avoid confusion

Sources/MlemMiddleware/FeedLoaders/Filtering/Filters/Post/PostKeywordFilter.swift
```diff
@@ -11,15 +11,15 @@ class PostKeywordFilter: FilterProviding {
...
     
-    init(keywords: Set<String>) {
-        self.keywords = keywords
+    init(context: FilterContext) {
+        self.context = context
     }
     
     func filter(_ targets: [Post2]) -> [Post2] {
-        let ret = targets.filter { !$0.title.lowercased().isContainedIn(keywords) }
+        let ret = targets.filter { shouldPassFilter($0) }
```
Sjmarf: *nit*
```suggestion
        let ret = targets.filter(shouldPassFilter)
```

Sjmarf (changes requested):

Sjmarf (approved):
LGTM! 💯

PR #88, opened 2024-12-27
Make all sent messages appear read

Sjmarf:
All sent messages are now considered read. In order to find out whether a message is our own, we need to fetch our own person ID and compare it to the message sender ID. I've implemented a system similar to the ApiClient `version` property for this - you can now `await api.myPersonId` to get the person ID.


EricBAndrews (approved):
LGTM! 🔥

PR #92, opened 2025-01-06
Efficient Subscription Check

EricBAndrews:
Adds a member to `SubscriptionList` containing the local ids of all subscribed communities, enabling efficient checking of subscription status in contexts where only that id is available (e.g., when only `Post1` is guaranteed to be available)


Sjmarf (approved):
LGTM 👍

PR #93, opened 2025-01-07
Prepend Item

EricBAndrews:
Makes `prependItem` a `StandardFeedLoader` and `FeedLoading` method


Sources/MlemMiddleware/FeedLoaders/Generics/LoadingActor.swift
```diff
@@ -66,6 +66,16 @@ actor LoadingActor<Item: FeedLoadable> {
...
     }
     
+    @discardableResult
+    func filterItem(_ target: Item) -> Item? {
+        let filtered = filter.filter([target])
+        if filtered.isEmpty {
+            return nil
+        } else {
+            return filtered[0]
+        }
```
Sjmarf: *nit* this could be simplified

```suggestion
        let filtered = filter.filter([target])
        return filtered.first
```

Sjmarf (approved):
LGTM 💯

PR #94, opened 2025-01-10
Add/Remove Administrators

EricBAndrews:
Adds logic for adding and removing administrators

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient+General.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient+General.swift
@@ -12,6 +12,17 @@ public extension ApiClient {
         myInstance?.administrators.contains(where: { $0.id == myPerson?.id }) ?? false
     }
     
+    /// Returns true if both myPerson and the given person are admins on this instance and myPerson outranks the given person, false otherwise
+    func isHigherAdmin(than person: any Person1Providing) -> Bool {
+        guard person.api.actorId == actorId,
+              let myPerson,
+              let myAdminIndex = myInstance?.administrators.firstIndex(of: myPerson.person2),
+              let targetAdminIndex = myInstance?.administrators.firstIndex(where: { $0.actorId == person.actorId } ) else {
+            return false
+        }
+        return myAdminIndex < targetAdminIndex
+    }
+    
     // Returns a raw API type :(
     // Probably OK because it's part of onboarding, which is cursed and bootstrappy
     func logIn(username: String, password: String, totpToken: String?) async throws -> ApiLoginResponse {
--- a/Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
@@ -56,18 +56,27 @@ public extension ApiClient {
 
 extension ApiClient {
     /// Adds or removes an admin from this API's instance
-    func addAdmin(_ person: any Person, added: Bool) async throws -> [Person2] {
-        let request = AddAdminRequest(personId: person.id, added: added)
+    @discardableResult
+    func addAdmin(personId: Int, added: Bool) async throws -> [Person2] {
+        let request = AddAdminRequest(personId: personId, added: added)
         let response = try await perform(request)
         
-        // only need to update this manually if removing admin, otherwise handled by above caching logic
-        if !added, let person2 = person as? any Person2Providing {
-            person2.person2.isAdmin = false
-        }
-        
-        return await caches.person2.getModels(
+        let updatedAdministrators = await caches.person2.getModels(
             api: self,
             from: response.admins
         )
+        
+        // update person's admin status
+        // only need to do this manually if removing admin, otherwise handled by above caching logic
+        if !added, let person = caches.person2.retrieveModel(cacheId: personId) {
+            person.isAdmin = false
+        }
+        
+        // update instance admins
+        if let myInstance {
+            myInstance.administrators = updatedAdministrators
+        }
+        
+        return updatedAdministrators
     }
 }
--- a/Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
@@ -53,3 +53,22 @@ public extension ApiClient {
         }
     }
 }
+
+extension ApiClient {
+    /// Adds an admin to this API's instance
+    func addAdmin(to instance: Instance3, personId: Int, added: Bool) async throws {
+        guard instance.api == self else {
+            assertionFailure("Instance API does not match self")
+            throw ApiClientError.invalidInput
+        }
+        
+        // the `to: instance` is a bit clunky but needed to set administrators without re-fetching the whole instance or
+        // asserting the presence of `myInstance`
+        let request = AddAdminRequest(personId: personId, added: added)
+        let response = try await perform(request)
+        instance.administrators = await caches.person2.getModels(
+            api: self,
+            from: response.admins
+        )
+    }
+}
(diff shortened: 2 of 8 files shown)


Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
```diff
@@ -53,3 +53,29 @@ public extension ApiClient {
...
+        }
+        
+        // the `to: instance` is a bit clunky but needed to set administrators without re-fetching the whole instance or
+        // asserting the presence of `myInstance`
+        let request = AddAdminRequest(personId: person.id, added: added)
+        let response = try await perform(request)
+        instance.administrators = await caches.person2.getModels(
+            api: self,
+            from: response.admins
+        )
```
Sjmarf: I don't think we need to require `instance` as an argument here. Would something like this work?

```suggestion
        let request = AddAdminRequest(personId: person.id, added: added)
        let response = try await perform(request)
        if let myInstance {
            myInstance.administrators = await caches.person2.getModels(
                api: self,
                from: response.admins
            )
        }
```

Sources/MlemMiddleware/API Client/ApiClient+General.swift
```diff
@@ -12,6 +12,17 @@ public extension ApiClient {
...
     
+    /// Returns true if both myPerson and the given person are admins on this instance and myPerson outranks the given person, false otherwise
+    func isHigherAdmin(than person: any Person1Providing) -> Bool {
+        guard person.api.actorId == actorId,
+              let myPerson,
+              let myAdminIndex = myInstance?.administrators.firstIndex(of: myPerson.person2),
+              let targetAdminIndex = myInstance?.administrators.firstIndex(where: { $0.actorId == person.actorId } ) else {
+            return false
+        }
+        return myAdminIndex < targetAdminIndex
```
Sjmarf, line 23: This relies on the presence of `myInstance`, which isn't ideal. If `myInstance` isn't present, it'll fail silently and return `false`. 

I'm happy to leave this as-is for now as it isn't particularly important. One potential way we could fix this would be to add an `administratorIds: [Int]` property to `ApiClient.Context`, and `await` that property inside of `isHigherAdmin`. Although, unlike the already existing `ApiClient.Context` properties, `administratorIds` would be liable to change after the context is first fetched. So, we'd have to make sure to keep it updated whenever the administrator list changes. That solution means having more than one source of truth for the admin list, which isn't ideal, but I think it's better than what we have now.

Sources/MlemMiddleware/Content Models/Person/Person1.swift
```diff
@@ -16,6 +16,7 @@ public final class Person1: Person1Providing {
     
     public let actorId: URL
     public let id: Int
+    public let local: Bool
```
Sjmarf: `Person` has an `apiIsLocal` computed property that returns whether it's local or not. Would it be possible to use that property instead?

Sjmarf (changes requested):

Sources/MlemMiddleware/Content Models/Person/Person1.swift
EricBAndrews: Sure, that works for me

Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
EricBAndrews: I've implemented something a little different--this function is only called from an `Instance3Providing`, so it now returns the `[Person2]` of the new admins, and it's up to the instance that calls it to update its own admin list. This approach also avoids unexpected behavior if `myInstance` isn't defined, though per the comment below `myInstance` is almost guaranteed to exist.

Sources/MlemMiddleware/API Client/ApiClient+General.swift
EricBAndrews: Hmm, that's a tricky one. I _really_ don't want to make this function `async` because it's called in `@ActionGroup`, which doesn't play too nice with `async`.

I think it's fine as-is because `myInstance` is _almost_ guaranteed to be defined--the `Session` initializers at the frontend are called at launch and account switch, they in turn call `getMyPerson`, and that sets `myInstance`, so unless that call fails (in which case we at least show an error toast) we can rely on `myInstance` being present.

Sjmarf: Ah, I see. That's fine by me, then 👍

Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
Sjmarf: > this function is only called from an Instance3Providing

Currently, every other update method in `ApiClient` (`voteOnPost`, `savePost` etc) is allowed to be called from outside of a middleware model and will update any existing middleware models itself. IMO we should follow that pattern here too, for consistency. We could update the `Instance3` model inside of this method by doing something like:

```swift
let administrators = caches.person2.getModels(
    api: self,
    from: response.admins
)
myInstance?.administrators = administrators
return administrators
```

This doesn't make the method dependent on the existence of `myInstance`. If an `Instance3` exists anywhere for the current `ApiClient`, it is guaranteed be in `myInstance` so we don't risk missing a state update by doing this.

Returning `[Person2]` is a good idea incase we need it, though it should be made `@discardableResult` if the method is changed to update the instance model.

Sjmarf: Also, it would be better for the function to accept `personId: Int` rather than `person: any Person` for consistency with other `ApiClient` methods

Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
```diff
@@ -56,18 +56,27 @@ public extension ApiClient {
 
 extension ApiClient {
     /// Adds or removes an admin from this API's instance
-    func addAdmin(_ person: any Person, added: Bool) async throws -> [Person2] {
```
Sjmarf: We can move this to the `public` extension above rather than being `internal`

Sources/MlemMiddleware/Content Models/Instance/Instance3Providing.swift
```diff
@@ -39,7 +39,7 @@ public extension Instance3Providing {
     var administrators_: [Person2]? { instance3.administrators }
     
     
-    func addAdmin(person: any Person, added: Bool) async throws {
-        instance3.administrators = try await api.addAdmin(person, added: added)
+    func addAdmin(personId: Int, added: Bool) async throws {
+        try await api.addAdmin(personId: personId, added: added)
     }
```
Sjmarf: In similar cases I've overloaded methods like this to also accept `Person` for convenience

```suggestion
    func addAdmin(personId: Int, added: Bool) async throws {
        try await api.addAdmin(personId: personId, added: added)
    }
    
    func addAdmin(_ person: any Person, added: Bool) async throws {
        addAdmin(personId: person.id, added: added)
    }
```

Sources/MlemMiddleware/Content Models/Instance/Instance3Providing.swift
Sjmarf, line 1: Does this need to be in `Instance3Providing` or could it be in `Instance1Providing`? Of course, in `Instance1Providing` we'd need to make sure to throw an error if it's not the local instance.

EricBAndrews: I can't think of a context where we'd need this available on `Instance1Providing`, since we're always calling it on `myInstance`. As you point out, it would be less safe on `Instance1`; the assertion would pretty easily handle that, but the fact that we'd need it inclines me towards leaving it on `Instance3Providing`. I don't mind either way.

Sjmarf: Sure, we can leave as-is then 👍

Sjmarf (approved):
LGTM!

PR #99, opened 2025-01-19
Search Comments

Sjmarf:


Sources/MlemMiddleware/FeedLoaders/Comment/SearchCommentFeedLoader.swift
```diff
@@ -0,0 +1,97 @@
...
+            nextCursor: nil
+        )
+    }
+}
+
+@Observable
+public class SearchCommentFeedLoader: StandardFeedLoader<Comment2> {
+    public var api: ApiClient
+    
+    // force unwrap because this should ALWAYS be a CommunityFetcher
```
EricBAndrews:
```suggestion
    // force unwrap because this should ALWAYS be a SearchCommentFetcher
```

Sources/MlemMiddleware/FeedLoaders/Generics/Fetcher.swift
```diff
@@ -109,7 +109,7 @@ public class Fetcher<Item: FeedLoadable> {
         cursor = nil
     }
     
-    func changeApi(to newApi: ApiClient, context: FilterContext) async {
+    public func changeApi(to newApi: ApiClient, context: FilterContext) async {
```
EricBAndrews: We solved this for `SearchPostFetcher` by overriding specifically on that fetcher--I'd prefer to keep that protection in place for non-search cases.

Subclassing `Fetcher` to a `SearchFetcher` that performs the public override would I think be the nicest way to handle this--we could also move the common parameters (query, communityId, creatorId, listing) into there and avoid some duplication. 

That would also require reworking `SearchPostFetcher` to be a `SearchFetcher` instead of `PostFetcher`, but I don't believe `SearchPostFetcher` actually leans on any of the `PostFetcher` behavior.

Sources/MlemMiddleware/FeedLoaders/Comment/SearchCommentFeedLoader.swift
```diff
@@ -0,0 +1,97 @@
+//
+//  File.swift
```
EricBAndrews:
```suggestion
//  SearchCommentFeedLoader.swift
```

EricBAndrews (changes requested):

Sources/MlemMiddleware/FeedLoaders/Generics/Fetcher.swift
Sjmarf: > We solved this for SearchPostFetcher by overriding specifically on that fetcher--I'd prefer to keep that protection in place for non-search cases.

Sure, I'll add an override instead 👍 

> Subclassing Fetcher to a SearchFetcher that performs the public override would I think be the nicest way to handle this--we could also move the common parameters (query, communityId, creatorId, listing) into there and avoid some duplication.

> I don't believe SearchPostFetcher actually leans on any of the PostFetcher behavior.

Yeah, that sounds like it would work well. At the moment the `SearchPostFeedLoader` inherits from `CorePostFeedLoader`, which requires its fetcher to be a `PostFetcher`. Performing that refactor would require us to shuffle things around quite a bit so I'd rather leave it out for now.

EricBAndrews (approved):
LGTM! 🚀

PR #101, opened 2025-01-22
Loops Integration

EricBAndrews:
Support for parsing loops.video content to embed in-app.

## Architecture Changes

To prefetch loops.video content, the media URL needs to be available before the prefetching begins. Once the content has been fetched, the URL must remain available for the frontend to consume without erasing the original link URL. To support this, `Post1` has a new field, `embeddedMediaUrl: URL?`. When this is populated, the post type is `.embedded` rather than `.media`, which allows the frontend to easily display the embedded media and a link to the original host.

`PrefetchingConfiguration` now has the additional field `embedLoops`. When `Post2Providing.imageRequests` is called, if `embedLoops: true`, it first attempts to parse the loops.video media URL and populates `embeddedMediaUrl`. When the image requests are then generated, they prefetch `embeddedMediaUrl`. If the parsing fails, `embeddedMediaUrl` remains `nil` and the post is treated as a link.

## Implementation Notes

- loops.video doesn't have API support for fetching the media URL directly, so we fetch the loops.video page and extract the URL from the HTML.
- The above operation is `async`, which makes `imageRequests()` async. To support this, I've added the `CollectionConcurrencyKit` library, which provides `concurrentFlatMap`.
- The `image` post type has been renamed `media`

### DIFF
--- a/Sources/MlemMiddleware/Content Models/Post/Post1Providing.swift
+++ b/Sources/MlemMiddleware/Content Models/Post/Post1Providing.swift
@@ -24,7 +24,7 @@ public protocol Post1Providing:
     var title: String { get }
     var content: String? { get }
     var linkUrl: URL? { get }
-    var loopsMediaUrl: URL? { get }
+    var embeddedMediaUrl: URL? { get }
     var deleted: Bool { get }
     var embed: PostEmbed? { get }
     var pinnedCommunity: Bool { get }
@@ -54,7 +54,7 @@ public extension Post1Providing {
     var title: String { post1.title }
     var content: String? { post1.content }
     var linkUrl: URL? { post1.linkUrl }
-    var loopsMediaUrl: URL? { post1.loopsMediaUrl }
+    var embeddedMediaUrl: URL? { post1.embeddedMediaUrl }
     var deleted: Bool { post1.deleted }
     var embed: PostEmbed? { post1.embed }
     var pinnedCommunity: Bool { post1.pinnedCommunity }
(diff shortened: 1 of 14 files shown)


Sjmarf:
> loops.video doesn't have API support for fetching the media URL directly, so we fetch the loops.video page and extract the URL from the HTML.

Yeah, it seems that way from the website. I wonder whether their mobile app uses an actual API 🤔 I was going to find out, but I don't have a Loops account yet (their confirmation email system is currently really slow, so I can't log in). I'm happy to go with this solution for now, though. If I manage to work it out I'll let you know

Sources/MlemMiddleware/Content Models/Post/Post1Providing.swift
```diff
@@ -117,16 +119,29 @@ public extension Post1Providing {
 }
 
 public extension Post1Providing {
+    /// If this post links to loops.video, attempts to parse the underlying media url and set embeddedMediaUrl
+    func parseLoopEmbeds() async {
+        if let loopsUrl = await linkUrl?.parseEmbeddedLoops() {
+            Task { @MainActor in
```
Sjmarf: I'm pretty sure we need to `await` the result of this task (in theory) because this value is read synchronously just after the call to `parseLoopEmbeds`. In practice this task probably completes fast enough for it to not matter, but it's better to be safe I suppose

Sources/MlemMiddleware/Extensions/URL+Extensions.swift
```diff
@@ -33,6 +33,27 @@ public extension URL {
         return components.url!
     }
     
+    /// Attempts to extract the underlying loops.video media URL from this URL
+    /// - Returns: loops.video media URL if this is a loops.video url and the underlying URL was successfully parsed, nil otherwise
+    func parseEmbeddedLoops() async -> URL? {
+        guard host() == "loops.video" else { return nil }
```
Sjmarf, line 42: Loops is federated, right? In theory there could be other loops instances that we might want to handle in future. I'm not sure whether there's a good way of checking whether it's a loops instance (besides maintaining a hard-coded domain list) unfortunately. Happy to just support `loops.video` for now 👍

Sjmarf (changes requested):

Sources/MlemMiddleware/Extensions/URL+Extensions.swift
EricBAndrews: Yeah, hopefully once their API matures we'll have something like a `/site` endpoint we can hit, and also maybe a more robust way of getting the media URL.

I've added a `TODO: pending loops.video maturation` comment noting these and opened an issue to track it. Once their API goes open source we can revisit.

Sjmarf (approved):
LGTM! 🚀

PR #105, opened 2025-01-27
Actor ID Changes

### LINKED ISSUE #90: Investigate making `ActorIdentifiable.host` (and dependent properties) non-optional

Sjmarf:
This PR changes much of the structure around `actorId` to resolve some incorrect assumptions that were previously made.

> [!note]
> Relies on [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/16).

## Issue no. 1

Before this PR, each `...Stub` struct contained an `actorId: URL` property. However, we didn't necessarily always intialise a `...Stub` using the *true* `actorId` of an object - rather, we used *any* url pointing to that object instead. One example of this is the logic for handling Lemmy links, which simply took the given URL and put it into a `...Stub`. In this way, we were kinda misuing the term "actor ID".

This meant that comparing the `actorId` of a `...Stub` with the `actorId` of a higher-tier model would not necessarily return `true` if the two objects referred to the same post/comment etc.

To resolve this, I've renamed the `actorId` property of `...Stub` models to `url`. `...Stub` models no longer conform to `ActorIdentifiable`. 

Additionally, all non-stub uses of `actorId` (such as in `Post1` etc) now use the type `ActorIdentifier` rather than `URL`. `ActorIdentifier` is a simple wrapper around `URL`. This makes it more explicit where we are handling a true actor ID and where we're handling any URL.

## Issue no. 2

Previously, we had assumed that all Actor IDs were in one of the following formats:
- `https://example.com`
- `https://example.com/c/name`
- `https://example.com/u/name`
- `https://example.com/post/123`
- `https://example.com/comment/123`
- `https://example.com/private_message/123`

However, this isn't actually the case. Because Lemmy can federate with non-Lemmy servers, we can also encounter other formats. For example, Kbin/Mbin use this format for communities:

- `https://fedia.io/m/fedia`

And Misskey uses this format for users:

- `https://misskey.io/users/9h75uqwaa8`

Notice how the Misskey URL does not include the user's username - it uses an ID instead.

Before this PR, we had extracted certain information about an entity from it's actor ID. For example, `CommunityStub`  had a `name: String` computed property that read from the second path component of the URL. Because different softwares may have different actor ID formats, we can't actually rely on this. As such, I've removed those computed properties.

## Issue no. 3

`host` computed properties are now non-optional. Closes #90


Sources/MlemMiddleware/Content Models/ActorIdentifier.swift
```diff
@@ -0,0 +1,97 @@
...
+    public static func instance(host: String) -> Self {
+        var components = URLComponents()
+        components.scheme = "https"
+        components.host = host
+        return ActorIdentifier(url: components.url!)!
+    }
+    
+    public var host: String {
+        url.host()! // We can force-unwrap here because we ensured that the host is extractable in `init`
+    }
```
EricBAndrews: *nit*

The "canon Swift" way to do this would be to store it as a property and set at `init`

Sources/MlemMiddleware/Content Models/ActorIdentifier.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    public var hostUrl: URL {
+        var components = URLComponents()
+        components.scheme = "https"
+        components.host = host
+        return components.url! // This will always succeed
```
EricBAndrews, line 62: *nit*

As with `host`, we can avoid the unwrap by guarding that this succeeds in the initializer.

I agree that the force unwrap is safe as used here, but this is such a fundamental piece of our middleware that I see no harm in an excess of caution, even though it costs a couple bytes per struct (especially since the initializer can already fail, so it doesn't have downstream contract problems)

Sources/MlemMiddleware/Content Models/ActorIdentifier.swift
```diff
@@ -0,0 +1,97 @@
...
+        guard url.host() != nil else { return nil }
+        self.url = url
+    }
+    
+    public static func instance(host: String) -> Self {
+        var components = URLComponents()
+        components.scheme = "https"
+        components.host = host
+        return ActorIdentifier(url: components.url!)!
+    }
```
EricBAndrews: From an API design perspective, it's better IMO to make this optional and move the force unwrap to the call point--we do need to force unwrap eventually, but it should be clear that this is technically an unsafe function when it is invoked.

EricBAndrews (changes requested):
Really nice work!

I've left a couple notes regarding force unwrapping. Given that we're not ingesting completely random URLs it's not a massive concern, but I still think we should avoid force unwraps where possible and avoid hiding them inside `ActorIdentifier`.

Sources/MlemMiddleware/Content Models/ActorIdentifier.swift
Sjmarf: That would work here, but wouldn't work where I've done the same thing in `URL+Extensions`:

```swift
    static func person(host: String, name: String) -> Self {
        var components = URLComponents()
        components.scheme = "https"
        components.host = host
        components.path = "/u/\(name)"
        return components.url! // This will always succeed
    }
 ```
 
 Because the documentation of `components.url` is so specific about when it returns `nil`, and we can guarantee that it will never do so, I don't think force unwrapping here is necessarily bad. The alternative would be to make the `person()` return type optional, which introduces all sorts of needless optional handling. Imo we should just use the force unwrap.

Sources/MlemMiddleware/Content Models/ActorIdentifier.swift
Sjmarf: I can refactor to avoid the force unwrap on `ActorIdentifier` 👍

See above comment for thoughts on the other unwrap

Sources/MlemMiddleware/Content Models/ActorIdentifier.swift
EricBAndrews: Works for me 👍

EricBAndrews (approved):
LGTM! 💯

PR #107, opened 2025-02-01
Mod Mail Feed Loader

EricBAndrews:
Implements feed loaders for mod mail.

The only weird thing here is that `ReportChildFeedLoader` itself uses a `MultiFetcher` of post/comment/message report child feed loaders; this is because the frontend displays all reports as a single feed, so it's ergonomic to have a single feed loader to drive that feed.

As far as I can tell, the new(ish) `FeedLoader` architecture with `MultiFetcher` means this sort of nesting just works (with a little bit of `init()` glue).

Other notes:
- I've added a `setup` method to `InboxFeedLoader` and `ModMailFeedLoader` that handles initializing all the child loaders for the default configuration and returns a big tuple of all of them
- Inbox items now use `inboxId` and `InboxIdentifiable` for identification instead of `ActorIdentifiable` to handle the fact that not all mod mail items are `ActorIdentifiable`. It's not _strictly_ necessary for the non-mod inbox items to use this, but down the line it may be convenient to have everything in the inbox be identifiable on a common field.


Sources/MlemMiddleware/FeedLoaders/Inbox/InboxFeedLoader/InboxItem.swift
```diff
@@ -0,0 +1,44 @@
...
+        switch self {
+        case let .message(message2): message2
+        case let .reply(reply2): reply2
+        }
+    }
+    
+    public var read: Bool {
+        switch self {
+        case .message(let message2): message2.read
+        case .reply(let reply2): reply2.read
```
Sjmarf: imo this syntax is a little nicer

```suggestion
        case let .message(message2): message2.read
        case let .reply(reply2): reply2.read
```

Sources/MlemMiddleware/FeedLoaders/Inbox/InboxFeedLoader/InboxItem.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+    
+    public var inboxId: Int {
+        var hasher: Hasher = .init()
+        
+        switch self {
+        case let .message(message): hasher.combine(message.actorId)
+        case let .reply(reply): hasher.combine(reply.actorId)
+        }
+
```
Sjmarf:
```suggestion
        hasher.combine(baseValue.actorId)
```

Sjmarf (approved):
Nice work! 🔥

PR #108, opened 2025-02-01
Fix instance actorId issue

Sjmarf:
Fixed an issue where instance actorIds sometimes had a trailing slash and sometimes didn't. This could cause comparisons to fail when they should have succeeded. For example, the instance block list wouldn't show any instances before this fix.

This doesn't require an Mlem PR


EricBAndrews (approved):
LGTM! ✅

PR #116, opened 2025-02-12
ApiClient hash value changes

Sjmarf:
Made some changes to the ApiClient architecture.

Previously, the `cacheId` and `hashValue` of `ApiClient` were created by hashing the `baseUrl` and the `token`. This wasn't ideal for a number of reasons:
- It was possible to create two different `ApiClient` instances for the same account by giving them different tokens.
- The `hashValue` of `ApiClient` would change if you change the token.
- It will be impossible to include `token` in `hashValue` if we ever switch to Swift 6 strict concurrency (issue [here](https://github.com/mlemgroup/mlem/issues/931)) and isolate `ApiClient` to an actor, because `token` is mutable.

To remedy this, I've added a new `username` property to `ApiClient`. If you want to create an authenticated `ApiClient`, you are required to provide the username of the account. `username` is now used instead of `token` in the `cacheId` and `hashValue`.

The downside of doing this is that we *must* have the `username` of an account available if we want to send requests using it. I can't think of a scenario in which we'd have a valid token but not the username though, so I think the benefits outweigh the risks.

--------

I've also made it possible to create a tokenless non-guest `ApiClient`. If you try to perform a request on such a client, it will throw `ApiClient.noToken`. This happens for all requests, including requests that don't require authentication.

This change makes the process of creating an authenticated `ApiClient` without knowing the token slightly easier. Previously, we had to do this:

```swift
let guestClient: ApiClient = .getApiClient(for: url, with: nil)
let loginResponse = try await guestClient.login(username: username, password: password)
let apiClient: ApiClient = .getApiClient(for: url, with: loginResponse.jwt)
```

After this change, you can now do this:

```swift
let apiClient: ApiClient = .getApiClient(url: url, username: username)
try await apiClient.login(password: password)
```

Mlem needs the guest `ApiClient` beforehand anyways to display the instance avatar in the login sheet, so I haven't changed the way in which Mlem authenticated clients for now. I may do some rewriting another time. It's still nice to have this syntax available though.

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -130,10 +130,10 @@ public class ApiClient {
     @discardableResult
     func perform<Request: ApiRequest>(
         _ request: Request,
-        allowedWhenTokenless: Bool = false // This should be `false` for the vast majority of requests, even GET requests
+        requiresToken: Bool = true // This should be `true` for the vast majority of requests, even GET requests
     ) async throws -> Request.Response {
         
-        guard allowedWhenTokenless || self.username == nil || self.token != nil else {
+        guard !requiresToken || self.username == nil || self.token != nil else {
             throw ApiClientError.noToken
         }
         
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -18,9 +18,11 @@ public class ApiClient {
     let decoder: JSONDecoder = .defaultDecoder
     let urlSession: URLSession = .init(configuration: .default)
     
-    // url and token MAY NOT be modified! Downstream code expects that a given ApiClient will *always* submit requests from the same user to the same instance.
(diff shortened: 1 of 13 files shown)


Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -109,28 +108,35 @@ public class ApiClient {
...
     @discardableResult
-    func perform<Request: ApiRequest>(_ request: Request) async throws -> Request.Response {
+    func perform<Request: ApiRequest>(
+        _ request: Request,
+        allowedWhenTokenless: Bool = false // This should be `false` for the vast majority of requests, even GET requests
+    ) async throws -> Request.Response {
+        
+        guard allowedWhenTokenless || self.username == nil || self.token != nil else {
+            throw ApiClientError.noToken
+        }
```
EricBAndrews: *nit*

Slightly nicer wording imo

```suggestion
        requiresToken: Bool = true // This should be `true` for the vast majority of requests, even GET requests
    ) async throws -> Request.Response {
        
        guard !requiresToken || self.username == nil || self.token != nil else {
            throw ApiClientError.noToken
        }
```

EricBAndrews (approved):
LGTM! 🚀

PR #120, opened 2025-02-16
v1 -> v2 upgrade

EricBAndrews:
Changes to support v1 -> v2 upgrade.

- `ApiClient.perform` now immediately throws `invalidToken` if token is "cannotRetrieveFromKeychain"
- `SharedTaskManager` now resets `ongoingTask` to nil when it completes. This prevents an issue where the task would be launched with an invalid API client, fail, but remain non-nil and therefore not try again.

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -139,10 +139,6 @@ public class ApiClient {
             throw ApiClientError.noToken
         }
         
-        guard token != "cannotRetrieveFromKeychain" else {
-            throw ApiClientError.invalidSession(self)
-        }
-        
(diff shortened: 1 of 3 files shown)


Sources/MlemMiddleware/API Client/ApiClient.swift
```diff
@@ -139,6 +139,10 @@ public class ApiClient {
             throw ApiClientError.noToken
         }
         
+        guard token != "cannotRetrieveFromKeychain" else {
+            throw ApiClientError.invalidSession(self)
+        }
```
Sjmarf: It would be better to avoid doing this by setting the token to `nil` in Mlem rather than `"cannotRetrieveFromKeychain"`. It's possible to do this as of #116

Sjmarf (changes requested):

Sjmarf (approved):
LGTM 👍

PR #121, opened 2025-02-17
Fix feed loaders sometimes not loading after applying filters

### LINKED ISSUE #1544: Inbox sometimes gets stuck loading infinitely when "unread only" is on
This is still an issue in some cases

EricBAndrews:
Closes mlemgroup/mlem#1544; does not strictly require a Mlem PR, but https://github.com/mlemgroup/mlem/pull/1767 prevents "load more" from flickering in for a frame.

Most of the line count is logging changes. The only code changes are StandardFeedLoader 146-148; `loadMoreItems()` is now automatically called when a filter is activated and the filtered items list is too short for threshold loading. This prevents the loader from entering an inescapable `idle` state.


Sjmarf (approved):
LGTM 🚀

PR #123, opened 2025-02-20
Tweak `ApiGetRequest` structure

### LINKED ISSUE #117: Make `...Request` types codable
Making the request types simple `Codable`-conforming structs would be nicer I think. We could then create some sort of custom `JSONEncoder` that creates the parameter list from the request

Sjmarf:
Changes the way that `ApiGetRequest` is structured. 

Here's what we did before:

```swift
public struct GetCommentRequest: ApiGetRequest {
    public typealias Response = ApiCommentResponse

    public let path = "comment"
    public let queryItems: [URLQueryItem]

    init(id: Int) {
        self.queryItems = [
            .init(name: "id", value: String(id))
        ]
    }
}
```

And here's what it looks like after this PR:

```swift
public struct GetCommentRequest: ApiGetRequest {
    public typealias Parameters = ApiGetComment
    public typealias Response = ApiCommentResponse
    
    public let path = "comment"
    public let parameters: Parameters?
    
    init(id: Int) {
        self.parameters = .init(id: id)
    }
}
```

This system brings the `GET` request structure closer in line with that of our `POST` requests, which use a `Body` value in a similar way to the `Parameters` value shown above. This structure makes it much easier to modify the parameters of an `ApiGetRequest` *after* initialising it, should we ever need to do that.

In order to encode the `parameters` into `[URLQueryItem]`, I've implemented a custom `URLQueryItemEncoder`.

Closes #117

See also [this lemmy-swift-codegen PR](https://github.com/mlemgroup/lemmy-swift-codegen/pull/12) and [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/17). No Mlem PR required.

### DIFF
--- a/Sources/MlemMiddleware/API Client/ApiClient.swift
+++ b/Sources/MlemMiddleware/API Client/ApiClient.swift
@@ -207,7 +207,7 @@ public class ApiClient {
     ) throws -> URLRequest {
         let token = tokenOverride ?? self.token
         guard permissions != .none else { throw ApiClientError.insufficientPermissions }
-        let url = definition.endpoint(base: endpointUrl)
+        let url = try definition.endpoint(base: endpointUrl, version: fetchedVersion ?? .infinity)
         var urlRequest = URLRequest(url: url)
         for header in definition.headers {
             urlRequest.setValue(header.value, forHTTPHeaderField: header.key)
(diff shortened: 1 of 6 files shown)


EricBAndrews (approved):
LGTM! 🚀

PR #124, opened 2025-02-22
v4 Api support

Sjmarf:
> [!important]
> This PR relies on [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/18). See also the [lemmy-swift-codegen PR](https://github.com/mlemgroup/lemmy-swift-codegen/pull/13).

Updated MlemApiTypes to include Lemmy 1.0, and made the minimum changes to MlemMiddleware required to support the updated Api types. The v4 API isn't actually called anywhere yet.


EricBAndrews (changes requested):
Nice work!

Something's not quite right with the API types--when I try to login to voyager.lemmy.ml, it throws a decoding error on the `GetSiteResponse`. I believe the discrepancy is due to the presence of the `oauth_providers` field in the response, which is not represented in our API type.

I'm also getting this decoding error on launch after upgrading from latest dev:
```
Optional(Swift.DecodingError.keyNotFound(CodingKeys(stringValue: "ap_id", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "communities", intValue: nil), _CodingKey(stringValue: "Index 1", intValue: 1), _CodingKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "value", intValue: nil), CodingKeys(stringValue: "apiCommunityView", intValue: nil), CodingKeys(stringValue: "community", intValue: nil)], debugDescription: "No value associated with key CodingKeys(stringValue: \"ap_id\", intValue: nil) (\"ap_id\").", underlyingError: nil)))
```

Sjmarf:
> Something's not quite right with the API types--when I try to login to voyager.lemmy.ml, it throws a decoding error on the GetSiteResponse.

It turns out that the backend version used by voyager.lemmy.ml doesn't currently have an equivalent `lemmy-js-client` tag. However, `lemmy-js-client` released another tag since I started working on this, which I've upgraded it to. It still doesn't perfectly match the API used by voyager.lemmy.ml, but it's close enough that logging in now succeeds.

Large parts of the app are broken on Lemmy 1.0. In addition to introducing the v4 api, they also made some *very* breaking changes to the v3 api. I'm hoping that's just temporary, because it kinda defeats the point of making a new API version. Some of those issues are kinda complicated so I won't be fixing them in this PR.

> I'm also getting this decoding error on launch after upgrading from latest dev:

This is caused by the recent search results file failing to decode. This happens because the recent search history was previously encoded in camel case, and now it's in snake case. Fixing this isn't particularly easy unfortunately, so I've silenced the error for now. This will cause users' search history to be wiped when they upgrade to this version.

EricBAndrews (approved):
LGTM! 🚀

PR #126, opened 2025-02-25
Prefetching refactor

Sjmarf:
Refactored prefetching logic.
- Move the prefetcher logic from `CorePostFeedLoader` to a generic `PrefetchingFeedLoader`
- Separated the prefetching for `Post1` and `Post2` into extensions of those respective classes

In 1.0 we'll need different types for the sort property of `SearchPostFeedLoader` and the other post feed loaders. At the moment all of the post feed loaders inherit from `CorePostFeedLoader`, which has a single sort type. Somehow it'll need to be refactored so that different post feed loaders can have different types for the `sort` property. I'm not sure how yet, but extracting the image prefetching into a separate class is probably the first step of that

This doesn't require an Mlem PR


EricBAndrews (approved):
LGTM! 🚀

PR #129, opened 2025-03-01
Improve keyword filtering logic

EricBAndrews:
Any non-letter character is now replaced with whitespace before keyword filters are applied. This catches a bunch of punctuation-related cases:
`keyword's`
`keyword-otherword`
`... keyword.`
`... keyword, ...`

Also renames `containsWordsIn` to `failsKeywordFilter`, since the logic is now pretty tightly coupled to that purpose.

### DIFF
--- a/Sources/MlemMiddleware/Extensions/String+Extensions.swift
+++ b/Sources/MlemMiddleware/Extensions/String+Extensions.swift
@@ -20,13 +20,9 @@ public extension String {
     
     /// Returns true if this string contains any whole word that is in the given set of strings
     func failsKeywordFilter(_ filteredKeywords: Set<String>) -> Bool {
-        // let punctuationRegex = "[^a-zA-Z]" // matches single non-letter characters
         let words = self
-            // .replacingOccurrences(of: punctuationRegex, with: " ", options: [.regularExpression])
-            // .split(separator: " ")
             .split(separator: /[^a-zA-Z]/) // split on any non-letter characters so "keyword's" is filtered as "keyword" "s"
             .map { $0.lowercased() }
-        print("DEBUG \(words)")
         return words.contains { filteredKeywords.contains($0) }
     }
 }
--- a/Sources/MlemMiddleware/Extensions/String+Extensions.swift
+++ b/Sources/MlemMiddleware/Extensions/String+Extensions.swift
@@ -20,11 +20,13 @@ public extension String {
     
     /// Returns true if this string contains any whole word that is in the given set of strings
     func failsKeywordFilter(_ filteredKeywords: Set<String>) -> Bool {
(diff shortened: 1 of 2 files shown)


Sources/MlemMiddleware/Extensions/String+Extensions.swift
```diff
@@ -19,8 +19,12 @@ public extension String {
...
     /// Returns true if this string contains any whole word that is in the given set of strings
-    func containsWordsIn(_ strings: Set<String>) -> Bool {
-        let words = self.split(separator: " ").map { $0.lowercased() }
-        return words.contains { strings.contains($0) }
+    func failsKeywordFilter(_ filteredKeywords: Set<String>) -> Bool {
+        let punctuationRegex = "[^a-zA-Z]" // matches single non-letter characters
+        let words = self
+            .replacingOccurrences(of: punctuationRegex, with: " ", options: [.regularExpression])
+            .split(separator: " ")
+            .map { $0.lowercased() }
```
Sjmarf: Iirc you can `.split` with a `RegexComponent`. Could do something like this which I think is a little cleaner

```swift
let words = self.split(separator: /[^a-ZA-Z0-9]+/).map { $0.lowercased() }
```

Sjmarf (approved):
Nice 👍

Sources/MlemMiddleware/Extensions/String+Extensions.swift
EricBAndrews: Nice, that's much cleaner 👍

Sources/MlemMiddleware/Extensions/String+Extensions.swift
```diff
@@ -19,8 +19,10 @@ public extension String {
     }
     
     /// Returns true if this string contains any whole word that is in the given set of strings
-    func containsWordsIn(_ strings: Set<String>) -> Bool {
-        let words = self.split(separator: " ").map { $0.lowercased() }
-        return words.contains { strings.contains($0) }
+    func failsKeywordFilter(_ filteredKeywords: Set<String>) -> Bool {
+        let words = self
+            .split(separator: /[^a-zA-Z]/) // split on any non-letter characters so "keyword's" is filtered as "keyword" "s"
```
d42ohpaz, line 24: It might be better to use `[^[:alnum:]]` (or `[:alpha:]` if you mean to ignore numbers) to include support for non-English languages. 

It is my understanding that Swift supports PCRE, so that syntax should work. Also, I would imagine `alnum` is better in case what I want to filter also contains numbers.

EricBAndrews: Thanks, that's a super useful suggestion!

PR #130, opened 2025-03-02
Improved Filtering 2

EricBAndrews:
Improves the keyword filtering regex to use `:alnum:` instead of `a-zA-Z` for the splitter; this should handle non-ASCII much more effectively.

Credit to [d42ohpaz](https://github.com/mlemgroup/MlemMiddleware/pull/129#discussion_r1976520427).

This does not require a Mlem PR.


Sjmarf (approved):
LGTM 👍

PR #131, opened 2025-03-04
Fix `not_a_mod_or_admin` errors

EricBAndrews:
Backend changes to address mlemgroup/mlem#1825. Does not require a Mlem PR, but [this Mlem PR](https://github.com/mlemgroup/mlem/pull/1826) is pointed at this branch so the changes can be easily tested.

Report feed loaders and the report count in `UnreadCount` now catch `not_an_admin` and `not_a_mod_or_admin` errors and return safe values.

Thanks to the joy of multithreading, there are some race conditions that cause these functions to be called from the frontend which would be extremely painful to resolve; the frontend doesn't ever display the empty responses unless something goes _seriously_ wrong, but the calls still get made and the errors still pop up.

### DIFF
--- a/Sources/MlemMiddleware/FeedLoaders/Inbox/ModMailFeedLoader/PostReportChildFeedLoader.swift
+++ b/Sources/MlemMiddleware/FeedLoaders/Inbox/ModMailFeedLoader/PostReportChildFeedLoader.swift
@@ -8,12 +8,16 @@
 public class PostReportChildFeedLoader: ModMailChildFeedLoader {
     class Fetcher: ModMailFetcher {
         override func fetchPage(_ page: Int) async throws -> FetchResponse {
-            let response = try await api.getPostReports(page: page, limit: pageSize)
-            return .init(
-                items: response.map { .report($0) },
-                prevCursor: nil,
-                nextCursor: nil
-            )
+            do {
+                let response = try await api.getPostReports(page: page, limit: pageSize)
+                return .init(
+                    items: response.map { .report($0) },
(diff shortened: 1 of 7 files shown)


Sources/MlemMiddleware/FeedLoaders/Inbox/ModMailFeedLoader/PostReportChildFeedLoader.swift
```diff
@@ -8,12 +8,16 @@
...
-            )
+            do {
+                let response = try await api.getPostReports(page: page, limit: pageSize)
+                return .init(
+                    items: response.map { .report($0) },
+                    prevCursor: nil,
+                    nextCursor: nil
+                )
+            } catch let ApiClientError.response(response, _) where response.error == "not_a_mod_or_admin" {
+                return .init(items: .init(), prevCursor: nil, nextCursor: nil)
```
Sjmarf: In the `ApiErrorResponse` file we've got some precedent for how we match these errors. It would be better to continue that precedent here, I think.

In the longer-term it would be nice to create an `enum` of the error types. This looks possible to do in lemmy-swift-codegen - all of the error types are listed in `LemmyErrorType.ts`.

Sjmarf (approved):
LGTM 👍

We might want to consider upgrading to Swift 6 strict concurrency mode (#931) at some point. Having the project in that mode eliminates the possibility of data races occurring. Looks quite time consuming though. It might amount to just putting `@MainActor` everywhere lol.

PR #132, opened 2025-03-07
Languages

Sjmarf:
`Instance3` now stores the array of `allLanguages` as `[Locale.Language]` rather than `[Int]`

This is for v2.1


EricBAndrews (approved):
LGTM! 🔥

PR #136, opened 2025-03-17
Add middleware sort types

Sjmarf:
Previously we dealt directly with the raw `Api...` sort types (`ApiSortType` and `ApiCommentSortType`) throughout the entire system, including in Mlem. In this PR, I've introduced a new layer of middleware sort types (`PostSortType`, `CommentSortType` and `SearchSortType`) that are easier to work with. The raw API sort types are now only used inside `ApiClient`.

This change is necessary because, in Lemmy 1.0, the sorting system has changed significantly. This new middleware layer handles both the old and new sorting systems. This may be subject to further changes, of course, because Lemmy 1.0 is still in the alpha stage. I've not properly tested these changes on Lemmy 1.0 yet.

In Lemmy 1.0, search sorting uses an `ApiSearchSortType` rather than `ApiSortType`. `ApiSearchSortType` only contains cases for `new`, `old` and `top`. This means that, from Lemmy 1.0 onwards, it will no longer be possible to search for posts/comments using sort modes like "Hot" or "Active". [This is intentional](https://github.com/LemmyNet/lemmy/issues/5401).

Of course, we want to continue supporting searching with those sort types for instances using versions older than 1.0. I've updated `SearchPostFeedLoader` so that it can use either `SearchSortType` or `PostSortType`. Once we drop support for pre-1.0, the `PostSortType` support can be removed from `SearchPostFeedLoader`. This solution isn't perfect, but I think it's worth maintaining support for those older search types whilst we still can.

I've also made the above change for `SearchCommentFeedLoader`.

Searching for communities and users have the same issue - the "Controversial" sort mode is no longer available for those search types from Lemmy 1.0 onwards. In this case I *don't* think it's worth hanging onto support for "Controversial", so the respective feed loaders only support `SearchSortType`. **The "controversial" sort mode has been removed from community and user search, for all Lemmy versions, in this PR**.


EricBAndrews (approved):
LGTM! 🚀

